Issue #25, opened 2017-09-16
Add TypeScript type definitions


lukechilds:
Will enable autocompletion in supported text editors/IDEs.

Previous discussion in https://github.com/lukechilds/keyv/issues/22

To prevent definitions going out of sync we should either include some sort of test to validate the definitions or alternatively they be generated from our source (annotated comments?).

caseyWebb:
> To prevent definitions going out of sync we should either include some sort of test to validate the definitions or alternatively they be generated from our source (annotated comments?).

I saw the previous discussion, but by far the easiest way to do this is rewriting the source in TypeScript. [tsd](https://www.npmjs.com/package/tsd) can be used to validate, but doesn't really solve the problem of them going out of sync as the `-test-d.ts` also has to be updated.

If you've had any change of heart, I've got no problem doing the rewrite and configuring tsc.

lukechilds:
Thanks for the offer @caseyWebb ♥

I kinda figured that was the only maintainable solution.

Tbh I've had a little play with TypeScript and wasn't really sold on it.

I'm interested in the idea, but I have some worries:

1) You port this codebase to TypeScript, I don't like it, don't accept your PR, you feel like you've wasted your time.

2) You port this codebase to TypeScript, I like it and merge, you disappear, future maintenance becomes much harder (I'm already struggling to find the time) because I'm not as familiar with TypeScript.

There is also the issue that if we re-write this in TypeScript we should probably re-write all storage adapters in TypeScript too.

What are your thoughts?

caseyWebb:
On the concern of not merging, no harm no foul. This codebase will take 15 minutes, tops.

On the concern of maintainability, it's actually a pretty easy out in worst case scenario. Set the tsconfig.json target to "esnext", transpile, replace the source with the output JS. It will be identical with only the types missing.

I'm more than willing to work on the adapters as well, as while not crucial I do agree that they should all be done. Have you considered/would you be open to a monorepo?

lukechilds:
You are definitely selling it to me.

>Have you considered/would you be open to a monorepo?

I've not been a fan of them historically but this current setup with dependencies spread across repos is definitely not ideal. If you can demonstrate a better solution using a monorepo then I'd definitely be interested.

The maintenance issue is the only thing really worrying me, as I mentioned before I'm already struggling with time as is. Extra build steps and stuff I'm not familiar with, while probably simple to learn, is just extra time eating away at time I already don't have a lot of.

[...]

PR #30, opened 2017-10-07
add typescript typing

tolbon:
this is my fisrt declaration types so maybe buggy at now.

Work In Progress #25

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,13 @@
+declare class Keyv {
+
+    constructor(uri: string, opts?: Object);
+    constructor(opts?: Object);
+
+    get(key: string): any;
+
+    set(key: string, value: any, ttl?: number): any;
+
+    delete(key: string): void;
+
+    clear(): undefined;
+}


lukechilds:
Thanks for this.

Any idea how we could include a tests to check the TypeScript definition is still valid?

tolbon:
Sorry for my response time,

Need to search about that but it's seems difficult or strange but need research about that

lukechilds:
No problem, I'm really busy myself, there's no rush.

Yeah, I'm not aware of a way to test, but I think we need to add a test before merging this. Otherwise API changes will pass tests but the TyepScript defs will be broken.

kirillgroshkov:
To test that typings are still valid I would do:

1. Create a typingsTest.ts file that consumes all API methods (calls them in different ways).
2. Run `tsc` on this file that should transpile it to js. If typings are correct - there will be no error (0 exitcode), otherwise test failed and error message is produced from `tsc`.

chroventer:
@lukechilds 
Check out this guide on TypeScript-testing: [Link](https://github.com/avajs/ava/blob/master/docs/recipes/typescript.md)

Issue #39, opened 2017-12-01
Allow returning the raw DB object


lukechilds:
Something like:

```js
const keyv = new Keyv();
await keyv.set('foo', 'bar');

await keyv.get('foo');
// 'bar'
await keyv.get('foo', { raw: true });
// { value: 'bar', expires: null }
```

PR #42, opened 2017-12-09
Implemented return of raw object

dusansimic:
Return of raw object is implemented in get function as requested in
issue #39.

Ref: https://github.com/lukechilds/keyv/issues/39

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -49,7 +49,7 @@ class Keyv extends EventEmitter {
 		return `${this.opts.namespace}:${key}`;
 	}
 
-	get(key) {
+	get(key, opts) {
 		key = this._getKeyPrefix(key);
 		const store = this.opts.store;
 		return Promise.resolve()
@@ -62,6 +62,9 @@ class Keyv extends EventEmitter {
 				if (typeof data.expires === 'number' && Date.now() > data.expires) {
 					this.delete(key);
 					return undefined;
+				}
+				if (opts && opts.raw) {
+					return data;
 				}
 				return data.value;
 			});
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -63,6 +63,9 @@ class Keyv extends EventEmitter {
 					this.delete(key);
 					return undefined;
 				}
+				if (opts && opts.raw) {
+					return data;
+				}
 				return data.value;
```
lukechilds: Would be nice if this just used the ternary operator.

test/keyv.js
```diff
@@ -75,6 +75,44 @@ test.serial('.set(key, val, ttl) where ttl is "0" overwrites default tll option
 	tk.reset();
 });
 
+test.serial('.get(key) gets only the value', async t => {
```
lukechilds: You don't need to test this, it's tested in [`@keyv/test-suite`](https://github.com/lukechilds/keyv-test-suite)

test/keyv.js
```diff
@@ -75,6 +75,44 @@ test.serial('.set(key, val, ttl) where ttl is "0" overwrites default tll option
...
+	const store = new Map();
+	const keyv = new Keyv({ store, ttl: 200 });
+	await keyv.set('foo', 'bar');
+	t.is(await keyv.get('foo'), 'bar');
+	tk.freeze(startTime + 250);
+	t.is(await keyv.get('foo'), undefined);
+	tk.reset();
+});
+
+test.serial('.get(key, {raw: true}) get db object insted only value', async t => {
```
lukechilds: You only need to test for the raw object here, no need to check the TTLs too.

lukechilds:
Perfect thanks!

Sorry one more thing, could you update the API docs.

dusansimic:
This should be it, no problem.

Issue #40, opened 2017-12-01
Expose Iterables


lukechilds:
Add `.keys()`, `.values()` `.entries()` methods similar to the `Map` spec but instead they should return a Promise that resolves to an iterable.

Something like:

```js
const keyv = new Keyv();
await keyv.set('foo', 'bar');
await keyv.set('fizz', 'buzz');

const keys = await keyv.keys();
for (const key of keys) {
  console.log(key);
}
// 'foo'
// 'fizz'

const values = await keyv.values();
for (const value of values) {
  console.log(value);
}
// 'bar'
// 'buzz'

const entries = await keyv.entries();
for (const [key, value] of entries) {
  console.log(`${key} => ${value}`);
}
// 'foo => bar'
// 'fizz => buzz'
```

chocolateboy:
> they should return a Promise that resolves to an iterable.

This requires *all* of the values to be slurped (or synchronously slurpable) before each method resolves, potentially incurring a huge cost in time and memory. What if there are a million records in a remote backend — taking, say, a minute to retrieve, and chewing through a gig of (mobile) RAM — but the caller only needs to check the first two?

```javascript
async function any (keyv, predicate) {
    const entries = await keyv.entries() // XXX uh-oh

    for (const [key, value] of entries) {
        if (predicate(key, value)) {
            return true
        }
    }
}
```

It would be more efficient to wait for a result, then process it, then wait for another result, then process that etc. The usual way to do this is not with a [recordset](https://en.wikipedia.org/wiki/Recordset) (e.g. a [promise which returns an array](https://github.com/lukechilds/keyv/pull/43)) but with a [cursor](https://en.wikipedia.org/wiki/Cursor_(databases)) (e.g. an [async iterator](http://2ality.com/2016/10/asynchronous-iteration.html)):

```javascript
async function any (keyv, predicate) {
    for await (const [key, value] of keyv.entries()) {
        if (predicate(key, value)) {
            return true
        }
    }
}
```

While *consuming* them in this way requires ES2018 (or Babel or [TypeScript](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#async-iteration)), producing them does not (just as any library can spit out promises for consumption by async/await). And, of course, the desugared usage is always available as a fallback:

```javascript
async function any (keyv, predicate) {
    const it = keyv.entries()[Symbol.asyncIterator]()

    while (true) {
        const { done, value: pair } = await it.next()

        if (done) {
            return false
        }

[...]

PR #43, opened 2017-12-11
Implemented exposed iterables

dusansimic:
Implemented exposed iterables as suggested in #40. Also added a function
_removeKeyPrefix so it's easy to remove prefix from key when returning
it in an iterable. Tests are added too for all three functions(keys,
values, entries).

Ref: https://github.com/lukechilds/keyv/issues/40

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@ test/testdb.sqlite
 #npm5
 package-lock.json
 
+#yarn
+yarn.lock
+
 # Logs
 logs
 *.log
--- a/src/index.js
+++ b/src/index.js
@@ -49,6 +49,10 @@ class Keyv extends EventEmitter {
 		return `${this.opts.namespace}:${key}`;
 	}
 
+	_removeKeyPrefix(key) {
+		return key.replace(`${this.opts.namespace}:`, '');
+	}
+
 	get(key) {
 		key = this._getKeyPrefix(key);
 		const store = this.opts.store;
@@ -97,6 +101,45 @@ class Keyv extends EventEmitter {
 		const store = this.opts.store;
 		return Promise.resolve()
 			.then(() => store.clear());
+	}
+
+	keys() {
+		const store = this.opts.store;
+		return Promise.resolve()
+			.then(() => store.keys())
+			.then(keys => {
+				const data = [];
+				for (const key of keys) {
+					data.push(this._removeKeyPrefix(key));
+				}
+				return data;
+			});
+	}
+
+	values() {
+		const store = this.opts.store;
+		return Promise.resolve()
+			.then(() => store.values())
+			.then(values => {
+				const data = [];
+				for (const value of values) {
+					data.push(this.opts.deserialize(value).value);
+				}
+				return data;
+			});
+	}
+
+	entries() {
+		const store = this.opts.store;
+		return Promise.resolve()
+			.then(() => store.entries())
+			.then(entries => {
+				const data = [];
+				for (const entry of entries) {
+					data.push([this._removeKeyPrefix(entry[0]), this.opts.deserialize(entry[1]).value]);
+				}
+				return data;
+			});
 	}
 }
 
--- a/test/keyv.js
+++ b/test/keyv.js
@@ -75,6 +75,30 @@ test.serial('.set(key, val, ttl) where ttl is "0" overwrites default tll option 
 	tk.reset();
 });
 
+test.serial('.keys()', async t => {
+	const store = new Map();
+	const keyv = new Keyv({ store });
+	await keyv.set('foo', 'bar');
+	await keyv.set('fizz', 'buzz');
+	t.deepEqual(await keyv.keys(), ['foo', 'fizz']);
+});
+
+test.serial('.values()', async t => {
+	const store = new Map();
+	const keyv = new Keyv({ store });
+	await keyv.set('foo', 'bar');
+	await keyv.set('fizz', 'buzz');
+	t.deepEqual(await keyv.values(), ['bar', 'buzz']);
+});
+
+test.serial('.entries()', async t => {
+	const store = new Map();
+	const keyv = new Keyv({ store });
+	await keyv.set('foo', 'bar');
+	await keyv.set('fizz', 'buzz');
+	t.deepEqual(await keyv.entries(), [['foo', 'bar'], ['fizz', 'buzz']]);
+});
+
 test.serial('Keyv uses custom serializer when provided instead of json-buffer', async t => {
 	t.plan(3);
 	const store = new Map();


thgh:
Can I help to get this merged?

rubenmoya:
I'd love to see this merged, is there anything missing in the PR? Anything we can help with?

lukechilds:
I'm not sure this is the correct way to go about this.

@chocolateboy has some vary valuable feedback here: https://github.com/lukechilds/keyv/issues/40#issuecomment-444632656

What are your thoughts?

lukechilds:
@TehShrike also has some valuable input here: https://github.com/lukechilds/keyv/issues/36#issuecomment-452739855

dusansimic:
I think @chocolateboy is right about that. We should do async iteration over keys/values/entries. Should we use the fallback version of iterating over entries or use Babel and use the ES2018 syntax (`for await`).

Also, should the iteration work like [filter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter) function on arrays or should predicate break the iteration? Filter like approach would still have to iterate over all of the entries but it also seem a bit more natural to me. Any thoughts on this?

lukechilds:
>Should we use the fallback version of iterating over entries or use Babel and use the ES2018 syntax (for await).

I'm not familiar with the fallback version, could you provide a tiny code example? I'd like to avoid Babel if possible but if `for await` would be a huge reduction in code complexity then I'd be open to it.

>Also, should the iteration work like filter function on arrays or should predicate break the iteration? Filter like approach would still have to iterate over all of the entries but it also seem a bit more natural to me. Any thoughts on this?

Sorry, not too sure I understand what you mean about a predicate. We should just allow `.keys()`, `.values()`, `.entries()` style methods something like in https://github.com/lukechilds/keyv/issues/40#issue-278444793.

If the users want's to filter for something while they're looping, that's up to them to implement.

Also, have you thought about how we can ensure these methods are atomic across storage adapters: https://github.com/lukechilds/keyv/issues/40#issuecomment-452929583

dusansimic:
> I'm not familiar with the fallback version, could you provide a tiny code example? I'd like to avoid Babel if possible but if `for await` would be a huge reduction in code complexity then I'd be open to it.

This is @chocolateboy s code that can be used as fallback instead of ES2018. `for await` does make code more readable and shorter but in general, it's the same thing.
``` javascript
async function any (keyv, predicate) {
    const it = keyv.entries()[Symbol.asyncIterator]()

    while (true) {
        const { done, value: pair } = await it.next()

        if (done) {
            return false
        }

        if (predicate(...pair)) {
            return true
        }
    }
}
```
But still, this is what user implements so, we shouldn't care much. If I understand the comments correctly, our goal i just to return an iterator which user can use to iterate over elements.

> Also, have you thought about how we can ensure these methods are atomic across storage adapters: [#40 (comment)](https://github.com/lukechilds/keyv/issues/40#issuecomment-452929583)

For storage adapters, cursor could be used. I think all databases that storage adapters use have that feature (for SQL and MongoDB it's cursor and for Redis its scan function). Any thoughts on this?

chocolateboy:
> But still, this is what user implements so, we shouldn't care much.

That's right. Whether to use Babel/TypeScript or a JS engine which supports `for await` (which at this point is [everything but IE11/Edge](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of#Browser_compatibility)) is a decision  for the *consumer* of the API.

It doesn't need to concern backends unless they a) elect to use *async* generators (if not, `for await` [works with synchronous iterables](http://2ality.com/2017/12/for-await-of-sync-iterables.html), so a backend can just return an ES6 collection or a plain old array) and b) want to support legacy browsers, in which case the *generator* can either be transpiled or written in longhand.

chocolateboy:
Incidentally, for Keyv v3, it would probably make sense for the API to restrict itself to synchronous iterables, rather than documenting/supporting the async usage. Why? Because the current API (`get`, `set`, `delete`, `clear`) is synchronous.

If/when a move to asynchronous iterables is made, e.g. for v4, the current API should be upgraded as well, IMO, i.e. `get`, `set`, `delete`, `clear` etc. should all return promises.

In practice, this doesn't mean every backend must be inherently async. It just means a backend *can* be async. As with async iterables, the synchronous behavior can be thought of as a subset/special case of the async API.

There's some prior art for this transition/migration in [Greasemonkey v3.0 -> v4.0](https://www.greasespot.net/2017/09/greasemonkey-4-for-script-authors.html).

lukechilds:
@dusansimic

>But still, this is what user implements so, we shouldn't care much. If I understand the comments correctly, our goal i just to return an iterator which user can use to iterate over elements.

Ah sorry, I misunderstood you, yeah, there is no `.any()` method on map, it shouldn't be implemented inside Keyv. I think @chocolateboy was giving an example of how user code would interact with the `.keys()`, `.values()` and `.entries()` Keyv methods.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map

>For storage adapters, cursor could be used. I think all databases that storage adapters use have that feature (for SQL and MongoDB it's cursor and for Redis its scan function). Any thoughts on this?

Yeah, the issue is making sure it works atomically, or if not possible across all storage adapters, then consistently. e.g, if they start looping over `.entries()` and then a new key gets inserted to the DB during the iteration, what happens? Does the entry get missed because it wasn't there at the beginning? Does it get missed or included depending on where the user is in the iteration when the key was inserted?

The specific behaviour isn't too important, what's important is that it's consistent across storage adapters.

@chocolateboy 

>Incidentally, for Keyv v3, it would probably make sense for the API to restrict itself to synchronous iterables, rather than documenting/supporting the async usage. Why? Because the current API (get, set, delete, clear) is synchronous.
>
>If/when a move to support asynchronous iteration is made, e.g. for v4, the current API should be upgraded as well, IMO, i.e. get, set, delete, clear etc. should all return promises.
>
>In practice, this doesn't mean every backend must inherently be async. It just means a backend can be async. As with async iterables, the synchronous behavior can be thought of as a trivial subset/special case of the async API.

That's already hoy Keyv works. It's already fully async, notice the `await` in the readme example:

```js
const Keyv = require('keyv');

// One of the following
const keyv = new Keyv();
const keyv = new Keyv('redis://user:pass@localhost:6379');
const keyv = new Keyv('mongodb://user:pass@localhost:27017/dbname');
const keyv = new Keyv('sqlite://path/to/database.sqlite');
const keyv = new Keyv('postgresql://user:pass@localhost:5432/dbname');
const keyv = new Keyv('mysql://user:pass@localhost:3306/dbname');

// Handle DB connection errors
keyv.on('error', err => console.log('Connection Error', err));

await keyv.set('foo', 'expires in 1 second', 1000); // true
await keyv.set('foo', 'never expires'); // true
await keyv.get('foo'); // 'never expires'
await keyv.delete('foo'); // true
await keyv.clear(); // undefined
```

https://github.com/lukechilds/keyv#usage

chocolateboy:
@lukechilds

> That's already how Keyv works. It's already fully async

Ah, great! I guess the [API documentation](https://github.com/lukechilds/keyv#instance) just needs updating:

- .set(key, value, [ttl])

> Returns true.

- .get(key, [options])

> Returns the value.

- .delete(key)

> Returns true if the key existed, false if not.

- .clear()

> Returns undefined.

lukechilds:
>Ah, great! I guess the API documentation just needs updating:

Woops, it does indeed, good spot!

officialpiyush:
Any updates on this?

dusansimic:
Hi @officialpiyush. I'm sorry that this issue is pr is inactive for such a long time. I didn't have enough time to fully invest in this and find a good solution.

The current state of my findings is that some database drivers that are used by keyv do not support exposed iterables and some don't even support cursors at all. The second problem which I didn't yet get to is checking if the behavior of drivers (iterables) is consistent across databases which is a deal breaker weather this will be implemented.

Right now I don't have much free time to work on this either so if someone else is interested in taking a shot at this, they're more than welcome. My suggestion would be that if it's needed, change the database driver used in the keyv storage adapters in order to have a better api and better features for work with this. It's far easier to just change the driver and tray to maintain consistency with the previous versions of keyv than try to work with inconsistent drivers.

thgh:
Do you have an overview of which adapters can/cannot handle this?

As I only need SQL adapters, I'm probably better off using knex?

dusansimic:
@thgh, currently no adapter is handling this, it needs to be implemented. As for sql database drivers, MySQL driver (mysql2) has some mentions of exposing iterables but I didn't test anything yet. Postgres driver (pg) has support for cursors and we might be able to implement an iterable with this. SQLite3 driver (sqlite3) does not support either iterables or any cursor implementation AFAIK. It either needs to be implemented using get method (return one entry) or we need to switch the database driver to for example `better-sqlite3` which seems to support iteration over results.

Knex is, if I understand correctly, a query builder, not a driver. Knex can build queries but it cannot execute them by itself, it uses the database driver behind the scenes.

Kikobeats:

PR #74, opened 2019-01-21
Implement delete multiple keys

dusansimic:
Passing array of keys to `.delete()` function deletes multiple keys all at once.

Closes: https://github.com/lukechilds/keyv/issues/41

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -46,6 +46,9 @@ class Keyv extends EventEmitter {
 	}
 
 	_getKeyPrefix(key) {
+		if (Array.isArray(key)) {
+			return key.map(key => `${this.opts.namespace}:${key}`);
+		}
 		return `${this.opts.namespace}:${key}`;
 	}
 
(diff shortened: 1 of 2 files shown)


dusansimic:
If this looks good, I'll just add it to docs and it's ready to pull.

src/index.js
```diff
@@ -90,7 +93,12 @@ class Keyv extends EventEmitter {
 		key = this._getKeyPrefix(key);
 		const store = this.opts.store;
 		return Promise.resolve()
-			.then(() => store.delete(key));
+			.then(() => {
+				if (store instanceof Map && Array.isArray(key)) {
```
lukechilds: This shouldn't only be implemented for Maps, the main benefit of using Keyv over alternatives is that it provides a consistent interface across multiple backends.

If we implement a feature only for a single backend it would break that.

lukechilds, line 97: This shouldn't only be implemented for Maps, the main benefit of using Keyv over alternatives is that it provides a consistent interface across multiple backends.

If we implement a feature only for a single backend it would break that.

test/keyv.js
```diff
@@ -75,6 +75,21 @@ test.serial('.set(key, val, ttl) where ttl is "0" overwrites default tll option
 	tk.reset();
 });
 
+test.serial('.delete(key) where key is a single key', async t => {
```
lukechilds, line 78: These tests should be in [`@keyv/test-suite`](https://github.com/lukechilds/keyv-test-suite) so they cover all storage adapters, not just Keyv + Map.

PR #75, opened 2019-01-22
Implement .has()

dusansimic:
Implementing `.has()` method from Map instances. Also adding `.isUndefined()` method since there's been a request for it in an issue referenced below.
The idea is to expose more raw data from Keyv via creating a method `.has()` that just tells us if the specified key exists and `.isUndefined()` method that tells us if that key has a value.

Closes: https://github.com/lukechilds/keyv/issues/50

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -86,14 +86,8 @@ class Keyv extends EventEmitter {
 			.then(() => true);
 	}
 
-	has(key) {
-		key = this._getKeyPrefix(key);
-		const store = this.opts.store;
-
-		return Promise.resolve()
-			.then(() => {
-				return Boolean(store.get(key, { raw: true }));
-			});
+	async has(key) {
+		return Boolean(await this.get(key, { raw: true }));
 	}
 
 	delete(key) {
--- a/src/index.js
+++ b/src/index.js
@@ -87,7 +87,7 @@ class Keyv extends EventEmitter {
 	}
 
 	async has(key) {
-		return Boolean(await this.get(key, { raw: true }));
+		return (await this.get(key, { raw: true })) !== undefined;
 	}
 
 	delete(key) {
--- a/src/index.js
+++ b/src/index.js
@@ -79,11 +79,24 @@ class Keyv extends EventEmitter {
 
 		return Promise.resolve()
 			.then(() => {
+				if (value === undefined) {
+					return false;
+				}
 				const expires = (typeof ttl === 'number') ? (Date.now() + ttl) : null;
 				value = { value, expires };
 				return store.set(key, this.opts.serialize(value), ttl);
 			})
-			.then(() => true);
+			.then(result => result !== false);
+	}
+
+	has(key) {
+		key = this._getKeyPrefix(key);
+		const store = this.opts.store;
+
+		return Promise.resolve()
+			.then(() => {
+				return Boolean(store.get(key, { raw: true }));
+			});
 	}
 
 	delete(key) {
--- a/src/index.js
+++ b/src/index.js
@@ -79,14 +79,11 @@ class Keyv extends EventEmitter {
 
 		return Promise.resolve()
 			.then(() => {
-				if (value === undefined) {
-					return false;
-				}
 				const expires = (typeof ttl === 'number') ? (Date.now() + ttl) : null;
 				value = { value, expires };
 				return store.set(key, this.opts.serialize(value), ttl);
 			})
-			.then(result => result !== false);
+			.then(() => true);
 	}
 
 	has(key) {
@@ -95,7 +92,17 @@ class Keyv extends EventEmitter {
 
 		return Promise.resolve()
 			.then(() => {
-				return Boolean(store.get(key, { raw: true }));
+				return store.has(key);
+			});
+	}
+
+	isUndefined(key) {
+		key = this._getKeyPrefix(key);
+		const store = this.opts.store;
+
+		return Promise.resolve()
+			.then(() => {
+				return store.has(key) && this.opts.deserialize(store.get(key)).value === undefined;
 			});
 	}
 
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -86,6 +86,26 @@ class Keyv extends EventEmitter {
 			.then(() => true);
 	}
 
+	has(key) {
+		key = this._getKeyPrefix(key);
+		const store = this.opts.store;
+
+		return Promise.resolve()
+			.then(() => {
+				return store.has(key);
```
lukechilds: You would need to implement the `.has()` method on all official storage adapters before this would work.

src/index.js
```diff
@@ -86,6 +86,26 @@ class Keyv extends EventEmitter {
...
+			});
+	}
+
+	isUndefined(key) {
```
lukechilds: I'm not sure an `.isUndefined()` method is necessary, it can be done outside of Keyv once `.has()` is implemented.

dusansimic:
> I'd suggest a better solution would be to allow passing an option to `get()` to return the raw value and checking if the raw value is set.

This is already implemented in https://github.com/lukechilds/keyv/pull/42.

lukechilds:
Woops sorry, just seen #42 is over a year old!

Apologies for the delay, it's been a very crazy year and have been struggling to keep up with OSS, will take a look at that PR now.

lukechilds:
@dusansimic I've reviewed https://github.com/lukechilds/keyv/pull/42 and made a few tweaks and merged.

If you could update this PR to make use of the raw option that would be brilliant.

Sorry again for the delay on that!

src/index.js
```diff
@@ -86,6 +86,16 @@ class Keyv extends EventEmitter {
 			.then(() => true);
 	}
 
+	has(key) {
+		key = this._getKeyPrefix(key);
+		const store = this.opts.store;
+
+		return Promise.resolve()
+			.then(() => {
+				return Boolean(store.get(key, { raw: true }));
```
lukechilds: This won't work, `store` references the underlying storage adapter which likely won't support the raw option.

src/index.js
```diff
@@ -86,6 +86,16 @@ class Keyv extends EventEmitter {
 			.then(() => true);
 	}
 
+	has(key) {
+		key = this._getKeyPrefix(key);
```
lukechilds: You don't need to re-implement `.get()` inside `.has()`, you can just call `.get()` inside it with `{ raw: true }`.

src/index.js
dusansimic: So it should be called via `this`?

src/index.js
dusansimic: Done

src/index.js
```diff
@@ -86,6 +86,10 @@ class Keyv extends EventEmitter {
 			.then(() => true);
 	}
 
+	async has(key) {
+		return Boolean(await this.get(key, { raw: true }));
```
lukechilds: Here I would probably check like this

```js
(typeof await this.get(key, { raw: true }) === 'object')
```

Although not sure that's necessarily better. Any specific reason you chose to wrap with `Boolean()`?

// CC @sindresorhus any input would be appreciated here 🙏

test/keyv.js
```diff
@@ -85,6 +85,14 @@ test.serial('.get(key, {raw: true}) returns the raw object instead of the value'
 	t.is(rawObject.value, 'bar');
 });
 
+test.serial('.has(key) where key is the key we are looking for', async t => {
```
lukechilds: These tests should actually be in [`keyv-test-suite`](https://github.com/lukechilds/keyv-test-suite/) so they are run against all storage adapters and catch any anomalies.

Don't worry about that for now, you can add the tests here and I'll move them over. The development process is a bit painful because all the modules depend on each other.

Would you be able to make the tests a bit more extensive though, it should test all possible different types of behaviour with all types. Ideally a single test for each check.

It may seem over the top and it's fine with the current implementation because you know it'll always return an object or undefined. But that's testing the current implementation, not the intended functionality.

If in the future we made a change to how entries are stored internally it could mean that a value of undefined would then cause `.has()` to return false when it should return true.

You should extensively test all expected behaviours with all types.

You can use these test files for inspiration:
https://github.com/lukechilds/keyv-test-suite/blob/master/src/api.js
https://github.com/lukechilds/keyv-test-suite/blob/master/src/values.js

src/index.js
dusansimic: Well it's just something I do so I would know I want to know if that value exists or not. If I use `typeof` it look to me like I'm checking for `object` specifically but what I actually want anything except `null`/`undefined`. I'd also like Sindres input on this.

lukechilds: > If I use typeof it look to me like I'm checking for object specifically but what I actually want anything except null/undefined.

That's exactly why I don't think it should be written like this. I think you **should** be checking for an object explicitly.

Using `{raw: true}` will return the raw keyv object if it's set or `undefined` if it's not. So you literally want to say "Does this function return an object." or "Does this function not return `undefined`"

So 

```js
typeof (await this.get(key, { raw: true })) === 'object'
```

or 

```js
(await this.get(key, { raw: true })) !== undefined
```

are semantically correct.

What you're currently doing:

```js
Boolean(await this.get(key, { raw: true }));
```

is just saying "Is this value truthy".

Which happens to work in this scenario, an object will be truthy and `undefined` will be falsey. However it's not checking the exact value we know it should be. It's less explicit and it's possible if we change the way data is structured internally it could break.

Does that make sense?

dusansimic:
I'm sorry for not finishing this up earlier, I had a lot of work around university. I've added all the requested changes and also created a PR on `keyv-test-suite` for the `has()` tests.

Kikobeats:

PR #79, opened 2019-03-08
API documentation: clarification that the Map-like instance methods return a promise

chocolateboy:
https://github.com/lukechilds/keyv/pull/43#issuecomment-471081525


Kikobeats:
Thanks for that!

PR #86, opened 2019-08-01
Add support for get() without key, returning all rows/documents in table/collection

tiagosiebler:
- db.get(key) works well when a key is provided, returning a specific row in the table.
- db.get() would work wonderfully to return all rows in a table, since no key was provided.

This PR makes it possible. Minor changes to make it happen:
- Moved isExpired logic to _checkIsExpired() fn, so it can be re-used per row, if multiple rows are returned.
- Added Array handler for _deserializeData() fn.

The only part I wasn't sure about was the name-spacing, should it return all rows in a particular name-space? Or all rows across all name-spaces?

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -4,52 +4,53 @@ const EventEmitter = require('events');
 const JSONB = require('json-buffer');
 
 const loadStore = opts => {
-	const adapters = {
-		redis: '@keyv/redis',
-		mongodb: '@keyv/mongo',
-		mongo: '@keyv/mongo',
-		sqlite: '@keyv/sqlite',
-		postgresql: '@keyv/postgres',
-		postgres: '@keyv/postgres',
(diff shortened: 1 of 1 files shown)


lukechilds:
Thanks for working on this.

This will need to have support in all official storage adapters before it can be merged. It will also need to have tests added in https://github.com/lukechilds/keyv-test-suite which run against all adapters.

Is this something you're willing to work on?

tiagosiebler:
I'll try when I can though I'm not sure yet when I'll find the time. If you or anyone else can help in the meantime, that'd be great of course. I've been using these changes in my implementation for a little while now and it's working well. Hopefully it's valuable to others too.

lukechilds:
Completely understand about not having time, I've been meaning to work on this for a while myself (https://github.com/lukechilds/keyv/issues/40) but also haven't had the time.

PR #89, opened 2019-08-29
Add goto-bus-stop/keyv-firestore third party adapter

Kikobeats:
Hello, I used this third party for creating a [self-hosted count service](https://github.com/Kikobeats/count) and it works like a charm, very happy with the result 🙂


lukechilds:
@Kikobeats nice work on count and good find!

@goto-bus-stop thanks for the sweet Firestore adapter!

PR #93, opened 2019-10-09
Refactor to Typescript + Monorepo

caseyWebb:
Need to do a proper writeup.

ref #25, #30 

tl;dr:

- existing packages are imported via lerna, hence the ridiculous commit history.

- yarn is required. This is one of the only times that it isn't a direct swap, because the repo is currently configured to use [workspaces](https://yarnpkg.com/lang/en/docs/workspaces/). This can be done with npm and lerna, but yarn workspaces are tremendously faster.

- scripts are exposed and intended to be executed from the top level. build, test, and lint support the `--watch` flag (e.g. `yarn build --watch`)

- lerna is only used for "release" command. if publishing manually from each directory is desired lerna can be yanked. That said, I highly recommend keeping it and pairing with conventional-commits+commitizen+commitlint to generate changelogs and automatically semver. [See here](https://github.com/Profiscience/knockout-contrib/blob/master/package.json)

- a note: I'm extremely strict with my TS and <3 generics. I've got a very strict linter configuration at the moment that might be too much of a PITA for someone who isn't well versed; I anticipate relaxing this after the initial refactor. The TS here shouldn't be taken as "I need to know how to do this".

### Todo
- [ ] scope coverage badges
- [ ] fix build status badges
- [x] port remaining @keyv/* packages
- [ ] add github templates for issues/prs


packages/keyv-mongo/src/index.ts
```diff
@@ -0,0 +1,112 @@
+import EventEmitter from 'events';
+import { MongoClient, Collection } from 'mongodb';
```
caseyWebb, line 2: `mongojs` was swapped for the native client which support promises ootb and has type definitions readily available

lukechilds: That's fine, as long as the test suit still passes.

lukechilds:
Thanks so much for putting this together. 😎

Ok, I've only had a quick skim through this, but the overall structure is looking very nice and much more manageable. I'm familiar with the concept of Lerna and workspaces but have not used them before so need to do a bit of reading there.

Docker Compose setup for testing is sweet af. Been meaning to get something like that working for a long time.

Seems like linting is killing the Travis builds for older Node.js versions. But the tests are still passing so that's good.

Also, do you know why the coverage has dropped so much? Keyv has 100% coverage, keyv-mongo has 98% and keyv-redis has 93%. So I'm not sure why it's dropped to 87% when combined to a monorepo. Maybe TS adding in extra fluff?

Also, I gotta be honest, some of this TypeScript syntax is pretty intimidating and not too readable for me. But that is something I expected as I'm pretty unfamiliar with it. I'll find some time to dig deeper and try and grok it a bit better.

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+	redis: '@keyv/redis',
+	mongodb: '@keyv/mongo',
+	mongo: '@keyv/mongo',
+	sqlite: '@keyv/sqlite',
+	postgresql: '@keyv/postgres',
+	postgres: '@keyv/postgres',
+	mysql: '@keyv/mysql'
+};
+
+const loadStore = <
```
caseyWebb, line 14: parameters in `<>` are [generics](https://www.typescriptlang.org/docs/handbook/generics.html). They're type params rather than runtime params essentially.

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+	TVal,
+	TAdapterOpts extends {
```
caseyWebb, line 16: using `extends` with a generic lets you require the type passed to satisfy constraints.

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+		adapter?: keyof typeof adapters;
+		uri?: string;
+	}
+>(opts: TAdapterOpts): KeyvStore<TVal> => {
```
caseyWebb, line 20: `: KeyvStore<TVal>` is an explicit return type which _forces_ the function to return something valid or throw an error. these are optional

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+		return new (adapter.__esModule ? adapter.default : adapter)(opts);
+	}
+
+	return new Map() as (Map<string, string> & { namespace: string });
+	/* eslint-enable @typescript-eslint/no-var-requires */
+};
+
+type MaybePromise<T> = T | Promise<T>;
+
+export interface KeyvOptions<TVal, TSerialized = string> {
```
caseyWebb, line 50: `TSerialized` is an optional generic with a default value of string

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+	namespace: string;
+	store: KeyvStore<TVal, TSerialized>;
+	adapter?: keyof typeof adapters;
```
caseyWebb, line 53: two operators going on here, `keyof` and `typeof`

`keyof` returns the value of the keys of an object. so `keyof { foo: string, bar: string }` is `'foo' | 'bar'`.

This could be a string, but would allow callers to pass invalid config. Good example of adding strictness that isn't strictly necessary but does serve a purpose aside from just verbosity.

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+	expires: number | null;
+	value: TVal;
+}
+
+export type KeyvStoreEvent = 'error';
+
+export default class Keyv<TVal> extends EventEmitter {
+	protected readonly opts: KeyvOptions<TVal>;
+
+	constructor(uriOrOpts: string | Partial<KeyvOptions<TVal>>, _opts: Partial<KeyvOptions<TVal>> = {}) {
```
caseyWebb, line 81: `Partial` is a ts builtin that takes a type that extends an object and makes every property optional

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+		super();
+
+		const adapterOpts = ({
+			namespace: 'keyv',
+			serialize: JSONB.stringify,
+			deserialize: JSONB.parse,
+			...((typeof uriOrOpts === 'string') ? { uri: uriOrOpts } : uriOrOpts),
+			..._opts
```
caseyWebb, line 89: The linter changed these to object rest spreads. Personally I prefer this syntax, but if you aren't familiar, the following are equivalent

```javascript
const foo = Object.assign({}, obj1, obj2)

const foo = {
  ...obj1,
  ...obj2
}
```

packages/keyv/src/index.ts
```diff
@@ -0,0 +1,160 @@
...
+		if (typeof this.opts.store.on === 'function') {
+			this.opts.store.on('error', err => this.emit('error', err));
+		}
+
+		this.opts.store.namespace = this.opts.namespace;
+	}
+
+	public async get(key: string, opts?: { raw: true }): Promise<void | KeyvStoreObject<TVal>>
+	public async get(key: string, opts?: { raw?: false }): Promise<void | TVal>
+	public async get(key: string, opts?: { raw?: boolean }): Promise<void | TVal | KeyvStoreObject<TVal>> {
```
caseyWebb, line 106: this is an overloaded method type that forces the return value based on the `raw` option. Benefits on both sides here as the caller will have the correct explicit return type, and the function can't return the wrong value if `raw` is true.

packages/keyv-mongo/src/index.ts
```diff
@@ -0,0 +1,112 @@
...
+	uri?: string;
+}
+
+interface Document<TVal> {
+	key: string;
+	value: KeyvStoreObject<TVal>;
+	expiresAt: Date | null;
+}
+
+export default class KeyvMongo<TVal> extends EventEmitter implements KeyvStore<TVal, KeyvStoreObject<TVal>> {
```
caseyWebb, line 17: `implements` forces the class to correctly implement the interface. in this case, it forces `get`, `set`, etc. to abide by the standard api for stores (`KeyvStore`)

packages/keyv-redis/src/index.ts
```diff
@@ -0,0 +1,89 @@
...
+	get(key: string): Promise<string>;
+	sadd(namespace: string, key: string): Promise<number>;
+	srem(namespace: string, key: string): Promise<unknown>;
+	smembers(namespace: string): Promise<string[]>;
+
+	set(key: string, value: string): Promise<unknown>;
+	set(key: string, value: string, mode: string, ttl: number): Promise<unknown>;
+}
+
+export interface KeyvRedisOptions extends redis.ClientOpts {
```
caseyWebb, line 17: this is fairly self-explanatory, but it adds the interface of redis's community supported types to this interface. lets consumers get auto-complete for redis options without needing to list them all and maintain them here.

lukechilds:
>Probably going to try to get it so that build/lint only run on the latest lts and tests run on all supported versions of node.

I was thinking the same thing, check out @bitcoinjs for some prior art on this:

https://github.com/bitcoinjs/bitcoinjs-lib/blob/41bf2cd03d85cd93a40992f5b9fea0af3e107734/.travis.yml

https://travis-ci.org/bitcoinjs/bitcoinjs-lib/builds/584039354?utm_source=github_status&utm_medium=notification

>I can go through and annotate the more confusing bits. I know some of it is pretty intense at first.

Thankyou, I really appreciate that 🙌

jaredwray:
Closing this down due to lack of response. If we need to open it back up makes sense. Looking in Q1 of 2021 to move this more to typescript and then look if we want to go mono repo or another architecture.

PR #94, opened 2019-11-01
build: use `buffer-json` instead of `json-buffer`

Kikobeats:
I noted `json-buffer` is an archived project, I replaced it for a `buffer-json`, it's basically the same but up to date and maintained.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ await cache.get('foo'); // 'cache'
 
 ### Custom Serializers
 
-Keyv uses [`json-buffer`](https://github.com/dominictarr/json-buffer) for data serialization to ensure consistency across different backends.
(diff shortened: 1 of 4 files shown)


lukechilds:
It doesn't look like the serialisation formats these two modules user are compatible.

Changing to `buffer-json` will cause invalid data to be returned for anyone who's already used a previous version of Keyv to store data as Buffers.

I agree `buffer-json` would be the better module to use but due to the fact that the old version is already widely in use, we'd have to implement some sort of database migration script to massage the different data formats over.

I think for this reason it's probably not worth the hassle.

Re being unmaintained, I don't think this is a huge issue, if you look at the source it's very simple and has no dependencies.

Thouhts?

Kikobeats:
👍

PR #95, opened 2019-11-01
feat: async serialize/deserialize support

Kikobeats:


lukechilds:
Can you add a test that checks async serialisers work?

The serialiser in the test can be something simple like:

```js
encode = data => Promise.resolve().then(() => JSON.stringify(data))
```

Kikobeats:
done 🙂

lukechilds:
Perfect, thanks!

PR #102, opened 2020-01-13
build: add `handleErrors` for more granular error control

### LINKED ISSUE #101: memory leak when passing a store
This line of code is attaching an event listener to a store passed as `store` option: https://github.com/lukechilds/keyv/blob/b87be6c60dd7ccf9efe8be3875c32c5b120fd7ff/src/index.js#L43

The thing is, a store can be shared between different keyv instances (the most common pattern there is reusing a Redis instance with difference namespace every time) so in that way the event listener is being added every time a keyv instance is created.

Kikobeats:
closes #101

> This line of code is attaching an event listener to a store passed as `store` option: https://github.com/lukechilds/keyv/blob/b87be6c60dd7ccf9efe8be3875c32c5b120fd7ff/src/index.js#L43
> 
> The thing is, a store can be shared between different keyv instances (the most common pattern there is reusing a Redis instance with difference namespace every time) so in that way the event listener is being added every time a keyv instance is created.
> 
>

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,7 +237,7 @@ A custom deserialization function.
 Type: `Boolean`<br>
 Default: `true`
 
-When it's `true`, errors on `options.store` will not be emitted at keyv level.
+When it's `true`, errors on `options.store` will be emitted at keyv level.
 
 #### options.store
 
--- a/README.md
+++ b/README.md
@@ -232,6 +232,13 @@ Default: `JSONB.parse`
 
 A custom deserialization function.
 
+#### options.handleErrors
+
+Type: `Boolean`<br>
+Default: `true`
+
+When it's `true`, errors on `options.store` will not be emitted at keyv level.
+
 #### options.store
 
 Type: `Storage adapter instance`<br>
--- a/src/index.js
+++ b/src/index.js
@@ -40,7 +40,7 @@ class Keyv extends EventEmitter {
 		}
 
 		if (typeof this.opts.store.on === 'function' && !this.opts.store.errorAlreadyHandled) {
-			this.opts.store.errorAlreadyHandled = true
+			this.opts.store.errorAlreadyHandled = true;
 			this.opts.store.once('error', err => this.emit('error', err));
 		}
 
--- a/src/index.js
+++ b/src/index.js
@@ -28,7 +28,8 @@ class Keyv extends EventEmitter {
 			{
 				namespace: 'keyv',
 				serialize: JSONB.stringify,
-				deserialize: JSONB.parse
+				deserialize: JSONB.parse,
+				handleErrors: true
 			},
 			(typeof uri === 'string') ? { uri } : uri,
 			opts
@@ -39,8 +40,7 @@ class Keyv extends EventEmitter {
 			this.opts.store = loadStore(adapterOpts);
 		}
 
-		if (typeof this.opts.store.on === 'function' && !this.opts.store.errorAlreadyHandled) {
-			this.opts.store.errorAlreadyHandled = true;
+		if (this.opts.handleErrors && typeof this.opts.store.on === 'function') {
 			this.opts.store.on('error', err => this.emit('error', err));
 		}
 
--- a/src/index.js
+++ b/src/index.js
@@ -41,7 +41,7 @@ class Keyv extends EventEmitter {
 
 		if (typeof this.opts.store.on === 'function' && !this.opts.store.errorAlreadyHandled) {
 			this.opts.store.errorAlreadyHandled = true;
-			this.opts.store.once('error', err => this.emit('error', err));
+			this.opts.store.on('error', err => this.emit('error', err));
 		}
 
 		this.opts.store.namespace = this.opts.namespace;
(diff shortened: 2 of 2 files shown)


lukechilds:
Don't you want each Keyv instance to be notified of an error though?

For example:

```js
const settingsCache = new Keyv({ store: redis });
settingsCache.on('error', handleSettingsCacheError); // Will fire on a DB error

const userCache = new Keyv({ store: redis });
userCache.on('error', handleUserCacheError); // Will never fire
```

lukechilds:
Also thinking about it, I'm not sure re-using a store is a supported use case.

I think you should create a new store instance for each Keyv instance. This is how it will work internally if you're just using connection URIs rather than passing store instances.

If you have some common settings you want to reuse you could create a factory function that returns a configured store instance you can pass into Keyv.

Kikobeats:
I think re-using a store is a common pattern in production scenarios.

For example, most apps are using a single SQL database and they are storing the data inside structured around tables. Also, from a code perspective, most of the time, you're keeping a singleton db instance since creating multiple instances for the same db connection is a waste of resources.

It's not a big deal to my use case and yea If I create multiple connections all is fine. Just I did this PR because it's an easy way to support more keyv use cases.

On the other hand, particularly on my use case, I set my redis instance to omit any kind of error (since I'm using it as an LRU cache and if an error happens I don't want to care), so actually this emitter will never emit something to me.

This is probably a very biased use case and it's necessary more use case inputs determine if it makes sense that it was something to support at the core level.

At least, I feel that establish a mechanism to prevent mutating the store is a worth it thing. Something like:

`{ handleErrors: false }`

Being `true` the default value, allowing customize it.

lukechilds:
> I think re-using a store is a common pattern in production scenarios.
> 
> For example, most apps are using a single SQL database and they are storing the data inside structured around tables. Also, from a code perspective, most of the time, you're keeping a singleton db instance since creating multiple instances for the same db connection is a waste of resources.

Ok yeah I guess that makes sense.

> It's not a big deal to my use case and yea If I create multiple connections all is fine. Just I did this PR because it's an easy way to support more keyv use cases.
> 
> On the other hand, particularly on my use case, I set my redis instance to omit any kind of error (since I'm using it as an LRU cache and if an error happens I don't want to care), so actually this emitter will never emit something to me.
> 
> This is probably a very biased use case and it's necessary more use case inputs determine if it makes sense that it was something to support at the core level.

I'm still a little concerned about breaking this behaviour: https://github.com/lukechilds/keyv/pull/102#issuecomment-574606063

I think that's probably a fairly common use case and I definitely want to remove any potential memory leaks from Keyv if possible.

Can you give me a little more insight into how you're using Keyv that causes the memory leak scenario?

I'm guessing just creating a few instances based on the same Redis store instance wouldn't cause any issues. So are you creating a new Keyv instance with a pre-existing Redis store instance on the fly each time a certain action triggers to end up with enough event emitters to create issues? Can you shine some light on that?

> At least, I feel that establish a mechanism to prevent mutating the store is a worth it thing. Something like:
> 
> { handleErrors: false }
> 
> Being true the default value, allowing customize it.

Yes this could be a solution. I'm not too happy with the way errors are handled currently in Keyv.

Kikobeats:
The thing I'm doing is reusing a redis instance where some options are configurable:

```js
const redisSingleton = require('./redis')

const createRedisCache = opts => {
  assert(opts.namespace, '`opts.namespace` is required.')
  assert(opts.ttl, '`opts.ttl` is required.')
  const store = REDIS_URI ? new KeyvRedis(redis) : new Map()
  return new Keyv({ store, ...opts })
}
```

lukechilds:
In your example you aren't re-using the storage adapter instance, you're creating a new storage adapter that uses the same `ioredis` instance.

Before your latest change using the `handleErrors` option you were adding/checking the `errorAlreadyHandled` property to the storage adapter instance not the `ioredis` instance. So with the above code example the `errorAlreadyHandled` property would always be false because Keyv always gets a fresh storage adapter instance.

Do you understand what I mean? Did you test the solution and did it stop the memory leak? As far as I can tell it wouldn't have changed the amount of event listeners that where created.

When and how does `createRedisCache()` get called in your application? Is this something that's happening a few times on app startup and regularly throughout the lifetime of your application?

README.md
```diff
@@ -232,6 +232,13 @@ Default: `JSONB.parse`
 
 A custom deserialization function.
 
+#### options.handleErrors
+
+Type: `Boolean`<br>
+Default: `true`
+
+When it's `true`, errors on `options.store` will not be emitted at keyv level.
```
lukechilds: This is the wrong way round, when it's `true` errors **will** be emitted.

Kikobeats:
Creating multiple storage adapter but reusing the store 🙂

After this, I'm not experimented the memory leak since I'm passing `handleErrors: false`}

lukechilds:
But just so I can gauge how the memory leak is getting triggered and if there's a better way to prevent this from happening in other use cases, how are you calling `createRedisCache()`?

Is this something that's happening a few times on app startup and regularly throughout the lifetime of your application?

szmarczak:
This PR doesn't solve the issue, as I see no point in having multiple `Keyv` instances attached to the same cache. There will be a memory leak as you can keep your amount of `Keyv` instances growing.

Kikobeats:
closing in favour of https://github.com/microlinkhq/keyv/pull/23

PR #107, opened 2020-05-04
Fix double prefix with TTL

soundstep:
Entries set with a TTL are not deleted, the `delete` call is made with a double prefix which does not delete the entry.

Corresponding issue: https://github.com/lukechilds/keyv/issues/108

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -51,10 +51,10 @@ class Keyv extends EventEmitter {
 	}
 
 	get(key, opts) {
-		key = this._getKeyPrefix(key);
+		const keyPrefixed = this._getKeyPrefix(key);
 		const { store } = this.opts;
 		return Promise.resolve()
-			.then(() => store.get(key))
+			.then(() => store.get(keyPrefixed))
 			.then(data => {
 				return (typeof data === 'string') ? this.opts.deserialize(data) : data;
 			})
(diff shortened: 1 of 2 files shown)


Kikobeats:
Hum, correct me if I'm wrong, the raw key need to be preserved in order definitely delete it at https://github.com/soundstep/keyv/blob/fix-double-prefix/src/index.js#L67, right?

BTW, @lukechilds can you take a look to this? 🙂

soundstep:
@Kikobeats Yeah both are needed, the param `key` remains to be sent the `delete` member function, which also do the prefix (that was the doubling), and the `keyPrefixed` is used to get the entry from the store (as before).

soundstep:
As an aside, re-assigning a function param is not great, ideally a `keyPrefixed` should be created every time the `_getKeyPrefix` is called. That would make things clearer, for example in the `set` function.
But that's "style" so I haven't done that in the PR.

lukechilds:
Thank you so much for this @soundstep, great spot!

lukechilds:
>As an aside, re-assigning a function param is not great, ideally a keyPrefixed should be created every time the _getKeyPrefix is called. That would make things clearer, for example in the set function.
But that's "style" so I haven't done that in the PR.

Can you explain a little bit more what you mean by this?

lukechilds:
Ahh I get it, implemented in https://github.com/lukechilds/keyv/commit/b87902889c0b8be123c3eee429ae55e6716d34ab

lukechilds:
Published as `keyv@4.0.1` 🎉

soundstep:
No problem @lukechilds !

PR #120, opened 2020-09-14
feature: allow empty namespace

vitormv:
### Why? 
* Basically I copy-pasted the code from this PR: https://github.com/lukechilds/keyv/pull/48, since it was stale for many months (thanks  @mjesun  for the original code and test)
* Also added the requested tests.
* Feature was also requested in this other PR https://github.com/lukechilds/keyv/issues/72

The main value of this change is that, when using most other adapters (sqlite, mysql, mongo, pgsql), its much cleaner to just store things in different tables/collections instead of using namespaces.

For example:
```js
const cache = new Keyv('mongodb://user:pass@localhost:27017/dbname', { collection: 'cache', namespace: null });
const users = new Keyv('mongodb://user:pass@localhost:27017/dbname', { collection: 'users', namespace: null });
```
or
```js
const keyv = new Keyv('sqlite://path/to/database.sqlite', { table: 'cache', namespace: null });
const users = new Keyv('sqlite://path/to/database.sqlite', { table: 'users' namespace: null });
```
The stored key remains pure, and helps with readability, and of course this is optional.

And huge thanks for the great project @lukechilds :D

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -47,7 +47,7 @@ class Keyv extends EventEmitter {
 	}
 
 	_getKeyPrefix(key) {
-		return `${this.opts.namespace}:${key}`;
+		return this.opts.namespace ? `${this.opts.namespace}:${key}` : key;
 	}
 
 	get(key, opts) {
--- a/src/index.js
+++ b/src/index.js
@@ -47,7 +47,7 @@ class Keyv extends EventEmitter {
 	}
 
 	_getKeyPrefix(key) {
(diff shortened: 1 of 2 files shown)


vitormv:
@lukechilds any opinions on this? 😄

jmealo:
@lukechilds: I hope you are well! I'm eager to know your opinions on this PR.

Kikobeats:

jaredwray:
With the removal of the namespace we dont know if we will completely not support that as it is built into every adapter.

PR #128, opened 2021-01-29
Update README.md

natesilva:
Add keyv-anyredis to the third-party storage adapters list. This client supports other Redis clients besides ioredis, and it also supports cluster mode for those clients that can use it (including ioredis).


Kikobeats:
Added at https://keyv.js.org 👍

PR #144, opened 2021-10-31
fix: typo in README

e0ipso:
There was an additional space in constructing a link.


jaredwray:
@e0ipso - thanks so much for the catch on this. Really appreciate it. 🎉

PR #189, opened 2022-01-04
Added KeyV-AzureTable Adapter

howlowck:
I created a Keyv adapter for [Azure Table](https://docs.microsoft.com/en-us/azure/cosmos-db/table/table-support). I'm hoping to add it to the list of third-party storage adapters for visibility.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,6 +147,7 @@ The following are third-party storage adapters compatible with Keyv:
 - [keyv-firestore ](https://github.com/goto-bus-stop/keyv-firestore) – Firebase Cloud Firestore adapter for Keyv
 - [keyv-mssql](https://github.com/pmorgan3/keyv-mssql) - Microsoft Sql Server adapter for Keyv
 - [keyv-memcache](https://github.com/jaredwray/keyv-memcache) - Memcache storage adapter for Keyv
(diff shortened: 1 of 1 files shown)


jaredwray:
@howlowck - it is great to see and thanks. Is there anything I can do to help on it?

howlowck:
@jaredwray Thanks for merging and for offering to help. Quick question, [is this the preferred way](https://github.com/howlowck/keyv-azuretable/blob/61f32dad129ead1b48d0e38841e8de4b0b36e4b1/src/index.ts#L91-L92) to emit/throw the error for `keyv`? (I mimicked from `keyv-memcache`) Thanks again!

jaredwray:
currently it is. I do believe long term there are better patterns to use. Do you have any suggestions?

howlowck:
Gotcha. Taking a quick glance at the `keyv` code, it looks like you are not catching the errors. This means that it's up to the user to catch them, which is totally a way to go if you are not changing/augmenting the error in `keyv`. My suggestion then is to provide examples in the README, instructing the users to catch the errors themselves. 

I'm not sure what the value of the 'error' event (via the `EventEmitter`) is; it looks like `keyv` isn't doing much with the event.

jaredwray:
Yep. current plan is to emit an error or throw based on how old the framework is. I am thinking of structuring it more moving forward.

Issue #166, opened 2020-08-18
MySQL - Urgent security issues in `lodash` and `sql`


julian-weinert:
I've just discovered (using `npm audit`) that there are some urgent and some minor issues.

• Package `sql` suffers from a not fixed SQL injection vulnerability (_not yet fixed_)
• Package `lodash` suffers from two urgent prototype pollution issues (_fixed in 4.17.11 and 4.11.17 respectively_)

Please consider updating `lodash` soon

jaredwray:
@julian-weinert and @identithree this has been resolved and is now published. There are no reliance on SQL or other packages moving forward on mysql. 🎉

PR #201, opened 2022-01-09
remove sqlite dependcies

alphmth:


packages/sqlite/package.json
```diff
@@ -1,59 +1,59 @@
 {
-  "name": "@keyv/sqlite",
```
jaredwray, line 2: Can you validate the formatting as we want to keep it inline with xo standards of this project. Looks like all of the package formatting has changed. You can do this by running `yarn xo` when in the package directory.

packages/sqlite/src/index.js
```diff
@@ -1,11 +1,14 @@
 'use strict';
 
-const KeyvSql = require('@keyv/sql');
+const EventEmitter = require('events');
 const sqlite3 = require('sqlite3');
 const pify = require('pify');
+const Sql = require('sql').Sql;
```
jaredwray: Is this package (https://www.npmjs.com/package/sql) up to date with security fixes? Looks like there has not been an update for over 4 years. Should we just go with normal SQL statements instead of introducing this package?

jaredwray:
@alphmth - looks like this package `Sql` https://github.com/brianc/node-sql is not well supported and we want to clean up our security issues with using this.

You can see the issue here where we want to move away from using that package: https://github.com/jaredwray/keyv/issues/166

jaredwray (approved):

PR #209, opened 2022-01-18
move to mongodb native driver

alphmth:


jaredwray:
Can you please add into tests the Official tests also:

```javascript
const { keyvOfficialTests } = require('@keyv/test-suite');

keyvOfficialTests(test, Keyv, mongoURL, 'mongodb://foo');
```

here is an example of what the test.js file should look like and work. 

```javascript
const test = require('ava');
const keyvTestSuite = require('@keyv/test-suite').default;
const { keyvOfficialTests } = require('@keyv/test-suite');
const Keyv = require('keyv');
const KeyvMongo = require('this');

const mongoURL = 'mongodb://127.0.0.1:27017';

keyvOfficialTests(test, Keyv, mongoURL, 'mongodb://foo');

const store = () => new KeyvMongo(mongoURL);
keyvTestSuite(test, Keyv, store);

test('default options', t => {
	const store = new KeyvMongo();
	t.deepEqual(store.opts, {
		url: mongoURL,
		collection: 'keyv',
	});
});

test('default options with url.uri', t => {
	const store = new KeyvMongo({ uri: mongoURL });
	t.is(store.opts.uri, mongoURL);
	t.is(store.opts.url, mongoURL);
});

test('Collection option merges into default options', t => {
	const store = new KeyvMongo({ collection: 'foo' });
	t.deepEqual(store.opts, {
		url: mongoURL,
		collection: 'foo',
	});
});

test('Collection option merges into default options if URL is passed', t => {
	const store = new KeyvMongo(mongoURL, { collection: 'foo' });
	t.deepEqual(store.opts, {
		url: mongoURL,
		collection: 'foo',
	});
});

test('.delete() with no args doesn\'t empty the collection', async t => {
	const store = new KeyvMongo('mongodb://foo'); // Make sure we don't actually connect
	t.false(await store.delete());
});

test('.delete() with key as number', async t => {
	const store = new KeyvMongo(mongoURL, { collection: 'foo' }); // Make sure we don't actually connect
	t.false(await store.delete(123));
});
```

This should validate namespace and other API issues also that you will need to get working.

jaredwray:
@alphmth - looks like you will need to dig in and understand why the tests are failing and adjust the updates to make them work. Did you see where there is an issue?

alphmth:
@jaredwray  MongoClient is not throwing errors so the test is not able to pass. I tried to fix it but it's not catching the error.

jaredwray:
@alphmth do you know what error should be thrown to make it pass?

alphmth:
@jaredwray its timeout issue. I have added `mongodb` property `serverSelectionTimeoutMS` and it starts to work

packages/mongo/src/index.js
```diff
@@ -28,10 +28,10 @@ class KeyvMongo extends EventEmitter {
 
 		// Implementation from sql by lukechilds,
 		this.connect = new Promise(resolve => {
-			mongoClient.connect(this.opts.url, { useNewUrlParser: true, useUnifiedTopology: true }
+			mongoClient.connect(this.opts.url, { useNewUrlParser: true, useUnifiedTopology: true, serverSelectionTimeoutMS: 5000 }
```
jaredwray: @alphmth - can we move this to be part of the options that are passed in?

jaredwray (changes requested):
Please move the serverSelectTimeout to be part of the options. In the tests you can pass in this value to get it passing.

jaredwray (approved):

PR #210, opened 2022-01-23
Convert sqlite3 to better-sqlite3

claabs:
[node-sqlite3](https://github.com/mapbox/node-sqlite3) appears to be somewhat abandoned and has been racking up security vulnerabilities in its dependencies. This PR switches it to [better-sqlite3](https://github.com/JoshuaWise/better-sqlite3) which is more performant, supports more architectures, and is actively maintained.

This PR also switches to using [Bind Parameters](https://github.com/JoshuaWise/better-sqlite3/blob/master/docs/api.md#binding-parameters) which resolves an issue I was experiencing when storing values that contain a single quote (`'`).

Note: I didn't really understand what was going on with the `TestAdapter`, so I just reverted [these changes](https://github.com/jaredwray/keyv/commit/ed094d0a0f7d492c5d0546dd5cd6cfe40f15708f#diff-a640cdc5ceede57b891b18b7ad8e7e093898041443596dde09b0b542fa9d9b87). The tests seem to still pass from what I see, although you may want to re-run manually to double check.

I understand this is a big change, so if you feel you want to stick with node-sqlite3 and just resolve its security vulnerabilities (like [this solution](https://github.com/mapbox/node-sqlite3/issues/1493#issuecomment-980521241)), I can just port this module into a standalone package. However, you still may want to consider switching to bind parameters, which node-sqlite3 also offers.

### DIFF
--- a/packages/keyv/test/test.js
+++ b/packages/keyv/test/test.js
@@ -1,11 +1,12 @@
-const EventEmitter = require('events');
 const test = require('ava');
-const keyvTestSuite = require('@keyv/test-suite').default;
-const { keyvOfficialTests } = require('@keyv/test-suite');
+const { keyvOfficialTests, default: keyvTestSuite } = require('@keyv/test-suite');
 const Keyv = require('this');
 const tk = require('timekeeper');
-const sqlite3 = require('sqlite3');
-const pify = require('pify');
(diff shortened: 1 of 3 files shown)


jaredwray:
@claabs - thanks so much for submitting this. The code looks great and we are moving it into the repo.

PR #218, opened 2022-01-27
support etcd

### LINKED ISSUE #55: Storage Adapter Etcd
Hi,
I'm writing the storage adapter for [Etcd](/coreos/etcd). Travis doesn't have an `etcd-service` available.
How would you manage to test it in CI?

alphmth:
Fixed #55


packages/etcd/package.json
```diff
@@ -0,0 +1,59 @@
+{
+	"name": "@keyv/etcd",
+	"version": "1.0.0",
+	"description": "Etcd storage adapter for Keyv",
+	"main": "src/index.js",
+	"scripts": {
+		"test": "xo --fix && nyc ava",
```
jaredwray: please remove the `--fix`

packages/etcd/src/index.js
```diff
@@ -0,0 +1,47 @@
+'use strict';
+
+const EventEmitter = require('events');
+const { Etcd3 } = require('etcd3');
+
+class KeyvEtcd extends EventEmitter {
+	constructor(url, options) {
+		super();
+		this.ttlSupport = false;
```
jaredwray: can we make this an option where you can set it too true? If ttlSupport is valid is there additional options needed to make it work?

packages/etcd/test/test.js
```diff
@@ -0,0 +1,37 @@
+const test = require('ava');
+const KeyvEtcd = require('this');
+
+const etcdURL = '127.0.0.1:2379';
+
```
jaredwray, line 14: Can you please add in the standard test-suite into this as it is used for all testing.

```javascript
keyvOfficialTests(test, Keyv, 'mysql://root@localhost/keyv_test', 'mysql://foo');

const store = () => new KeyvMysql('mysql://root@localhost/keyv_test');
keyvTestSuite(test, Keyv, store);
```

jaredwray (changes requested):

packages/etcd/test/test.js
stropicall: @jaredwray Can you help me in handling errors?  The documentation says to handle errors as specified in [link](https://microsoft.github.io/etcd3/interfaces/ioptions.html#faulthandling)

But Its not working

jaredwray: When you add in the code above does it fail locally? If so, have you tried to use a try/catch around the code and if it fails to handle it correctly?

jaredwray:
@alphmth - is this ready for review? Did you get all the changes in? If so, let me know when this is ready.

packages/etcd/test/test.js
stropicall: I have fixed it with help of this https://github.com/microsoft/etcd3/issues/105

alphmth:
@jaredwray namespace test is failing so working on that. hopefully by tomorrow you can review

jaredwray (approved):

Issue #198, opened 2022-01-08
unexpected undefine after I set the value


alsotang:
```js
const Keyv = require('keyv');
const path = require('path')

const keyv = new Keyv(`sqlite://${path.join(__dirname, 'database.sqlite')}`);

async function main() {
  let v = await keyv.get('foo')

  console.log(v)

  await keyv.set('foo', 'bar', 1000)

  v = await keyv.get('foo')

  console.log(v)
}

main()
```

![image](https://user-images.githubusercontent.com/1147375/148651916-34d75c9c-cb73-4606-b90a-ef3bace51fd2.png)

When `await keyv.set('foo', 'bar', 1000)` this line is executed, the `get` result should absolutely return a value but it is not.

You can debug the bug by running the script every second.

jaredwray:
@alsotang - will be looking into this over the next couple days to validate.

jaredwray:
@alsotang - can you verify this with the latest update as with sqlite this seems to be working correctly.

alsotang:
It's still broken after I cleared `node_modules` and reinstalled `keyv`

PR #219, opened 2022-01-27
wait for the detele promise

alsotang:
Because we don't know whether the `delete` would occur first or the next `set` operation. 

So just wait for the delete. Performance is not the point.

### DIFF
--- a/packages/keyv/src/index.js
+++ b/packages/keyv/src/index.js
@@ -62,8 +62,7 @@ class Keyv extends EventEmitter {
 				}
 
 				if (typeof data.expires === 'number' && Date.now() > data.expires) {
-					this.delete(key);
-					return undefined;
+					return this.delete(key).then(() => undefined);
 				}
 
 				return (options && options.raw) ? data : data.value;
--- a/packages/keyv/test/test.js
+++ b/packages/keyv/test/test.js
@@ -128,3 +128,53 @@ test.serial('Keyv supports async serializer/deserializer', async t => {
 	await keyv.set('foo', 'bar');
 	t.is(await keyv.get('foo'), 'bar');
 });
+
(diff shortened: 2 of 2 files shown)


jaredwray:
@alsotang - can you provide testing around this as what would happen if it does not come back? Is this backwards compatible across all of the system?

alsotang:
Yes, it's backward compatible because I think this is some kind of bug fix.

This pr is come from #198 . I think I figured out why my `set` op in the issue doesn't work.

Let's simulate two round of the reproduction code.

```js

// round 1
  let v = await keyv.get('foo') // return undefined. as expected

  console.log(v) // print undefined. as expected

  await keyv.set('foo', 'bar', 1000) // set success. as expected

  v = await keyv.get('foo') // return 'bar'. as expected

  console.log(v) // print 'bar'

// sleep 2s
// round 2
  let v = await keyv.get('foo') // found `foo` is expired, return undefined. and plan to detele the key in nearest future but not right now.

  console.log(v) // print undefined

  await keyv.set('foo', 'bar', 1000) // set success. 

 // now the delete code start to run. and the `set` op in the last line would be no effect

  v = await keyv.get('foo') // return undefined. not as expected(!!!)

  console.log(v) // print undefined
```

As I said, 

> we don't know whether the delete would occur first or the next set operation

If the `set` occurs first, then it would be detele by mistake.

jaredwray:
@alsotang - can you add a test in `https://github.com/jaredwray/keyv/blob/master/packages/keyv/test/test.js` to support this?

alsotang:
Maybe I can't. I could not find a specified database to always behave as I assumed. 

Could you understand what my example means?

alsotang:
ok, I wrote the tests. See what changed.

PR #224, opened 2022-02-01
fixed bigint and symbol

### LINKED ISSUE #220: Cannot store bigint and symbol values
### Problem
When using `Keyv#set`, an error is thrown if the `value` is a `symbol` or `bigint`, even though the documentation states that "Values can be of any type.". This also applies if the values are objects that contain one of those types.

For symbols, the error message is vague: 
```
SyntaxError: Unexpected token u in JSON at position 9
    at JSON.parse (<anonymous>)
    at Object.exports.parse [as deserialize] (C:\<project path>\node_modules\.pnpm\json-buffer@3.0.1\node_modules\json-buffer\index.js:49:15)
```

For bigints, the error message is more clear:
```
TypeError: Do not know how to serialize a BigInt
    at JSON.stringify (<anonymous>)
    at stringify (C:\<project path>\node_modules\.pnpm\json-buffer@3.0.1\node_modules\json-buffer\index.js:45:17)
    at Object.stringify [as serialize] (C:\<project path>\node_modules\.pnpm\json-buffer@3.0.1\node_modules\json-buffer\index.js:32:37)
    at C:\<project path>\node_modules\.pnpm\keyv@4.0.5\node_modules\keyv\src\index.js:89:22
    at async C:\<project path>\test-keyv-symbol.js:10:4
```

### How to reproduce
Run this demonstration code in a directory with Keyv added as a dependency:
```js
const Keyv = require('keyv');

const testDB = new Keyv('sqlite://test.sqlite', { namespace: 'test' });
testDB.on('error', err => console.log('Connection Error', err));

(async () => {
	const values = ['Normal string value', 1, true, Symbol('Symbol value'), 1n, { foo: Symbol('Nested symbol value') }, { [Symbol('Symbol key')]: 'bar' }];
	for (const value of values) {
		try {
			await testDB.set('key', value);
			console.log('Success: ', await testDB.get('key'));
		}
		catch (error) {
			console.log('Failure: ', error);
		}
	}
})();
```
The output will show `Success:` lines for the first 3 values (string, number, and boolean), but show `Failure:` for the next 3 values (symbol, bigint, and an object containing a symbol). For the final value (an object with a symbol key), the output will show `Success:  {}`, which is effectively a silent failure.

### Expected behavior
My understanding is that it's impossible to serialize and de-serialize a `symbol`, because they are unique by definition. As such, the user should be clearly and eagerly warned when attempting to store them. Rather than the vague `undefined` JSON error or silent failure, an error message like `TypeError: Symbols cannot be serialized.` should be emitted whenever the value is a symbol or an object containing a symbol as a key or value.

For bigints, these should be serialized and de-serialized normally, like other primitives. 

### Version
4.0.5

alphmth:
Fixes #220

### DIFF
--- a/packages/keyv/src/index.js
+++ b/packages/keyv/src/index.js
@@ -4,7 +4,9 @@ const EventEmitter = require('events');
 const JSONB = require('json-buffer');
 
 // eslint-disable-next-line no-extend-native
-BigInt.prototype.toJSON = function() { return this.toString()  }
+BigInt.prototype.toJSON = function () {
+	return this.toString();
+};
 
 const loadStore = options => {
 	const adapters = {
--- a/packages/keyv/src/index.js
+++ b/packages/keyv/src/index.js
@@ -3,6 +3,9 @@
 const EventEmitter = require('events');
 const JSONB = require('json-buffer');
 
+// eslint-disable-next-line no-extend-native
+BigInt.prototype.toJSON = function() { return this.toString()  }
+
 const loadStore = options => {
 	const adapters = {
 		redis: '@keyv/redis',
@@ -86,6 +89,10 @@ class Keyv extends EventEmitter {
 		return Promise.resolve()
 			.then(() => {
 				const expires = (typeof ttl === 'number') ? (Date.now() + ttl) : null;
+				if (typeof value === 'symbol') {
+					this.emit('error', 'symbol cannot be serialized');
+				}
+
 				value = { value, expires };
 				return this.opts.serialize(value);
 			})
--- a/packages/test-suite/src/values.js
+++ b/packages/test-suite/src/values.js
@@ -56,6 +56,24 @@ const keyvValueTests = (test, Keyv, store) => {
 		t.deepEqual(await keyv.get('foo'), value);
 	});
 
+	test.serial('value can not be symbol', async t => {
+		const keyv = new Keyv({ store: store() });
+		const value = Symbol('value');
+		try {
+			await keyv.set('foo', value);
+		} catch (error) {
+			t.is(error.context, 'symbol cannot be serialized');
+		}
+	});
+
+	test.serial('value can be bigint', async t => {
+		const keyv = new Keyv({ store: store() });
+		// eslint-disable-next-line no-loss-of-precision
+		const value = 9_223_372_036_854_775_807;
+		await keyv.set('foo', value);
+		t.deepEqual(await keyv.get('foo'), value);
+	});
+
 	test.after.always(async () => {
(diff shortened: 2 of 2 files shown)


packages/keyv/src/index.js
```diff
@@ -4,7 +4,9 @@ const EventEmitter = require('events');
 const JSONB = require('json-buffer');
 
 // eslint-disable-next-line no-extend-native
-BigInt.prototype.toJSON = function() { return this.toString()  }
+BigInt.prototype.toJSON = function () {
+	return this.toString();
```
stropicall, line 8: @jaredwray we need to exclude this line from coverage i think

packages/keyv/src/index.js
```diff
@@ -3,6 +3,11 @@
 const EventEmitter = require('events');
 const JSONB = require('json-buffer');
 
+// eslint-disable-next-line no-extend-native
+BigInt.prototype.toJSON = function () {
+	return this.toString();
```
glasser, line 8: This won't round-trip properly. Deserialization will yield a string rather than a bigint (typeof === 'string' not typeof === 'bigint').

packages/test-suite/src/values.js
```diff
@@ -56,6 +56,24 @@ const keyvValueTests = (test, Keyv, store) => {
...
+			await keyv.set('foo', value);
+		} catch (error) {
+			t.is(error.context, 'symbol cannot be serialized');
+		}
+	});
+
+	test.serial('value can be bigint', async t => {
+		const keyv = new Keyv({ store: store() });
+		// eslint-disable-next-line no-loss-of-precision
+		const value = 9_223_372_036_854_775_807;
```
glasser, line 72: This is not a bigint; this is just a normal JS number that's so large that it gets immediately rounded.

You can replace this line with
```
const value = BigInt('9223372036854775807');
```

and then the test will fail because of the issue I describe in my other comment.

If you want to be able to round-trip BigInts through JSON, you need an unambiguous way to represent them in the protocol, or a way of deciding when numbers should be BigInts.

My suggestion is that this change (which mutates the global BigInt, which is icky too) be rolled back, and that folks who want to use bigints supply a custom serializer (perhaps using https://www.npmjs.com/package/json-bigint or something).

stropicall: @glasser there is already an option to supply a custom serializer.

glasser: I understand; my point is that since "JSON encoding of bigints" is not particularly standardized, using a custom serializer might be more appropriate for handling them. (Alternatively, the default serializer could perhaps use something other than JSON that natively supports more types.)

PR #225, opened 2022-02-02
Expose Iterables

### LINKED ISSUE #40: Expose Iterables

alphmth:
Fixes #40

### DIFF
--- a/packages/keyv/src/index.js
+++ b/packages/keyv/src/index.js
@@ -57,6 +57,9 @@ class Keyv extends EventEmitter {
 					? iterator(this.opts.store.namespace)
 					: iterator) {
 					const data = typeof raw === 'string' ? this.opts.deserialize(raw) : raw;
+					if (this.opts.store.namespace && !key.includes(this.opts.store.namespace)) {
+						continue;
+					}
 
 					if (typeof data.expires === 'number' && Date.now() > data.expires) {
 						this.delete(key);
--- a/packages/keyv/src/index.js
+++ b/packages/keyv/src/index.js
@@ -57,9 +57,6 @@ class Keyv extends EventEmitter {
 					? iterator(this.opts.store.namespace)
 					: iterator) {
 					const data = typeof raw === 'string' ? this.opts.deserialize(raw) : raw;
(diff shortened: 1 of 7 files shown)


packages/keyv/src/index.js
```diff
@@ -73,7 +70,7 @@ class Keyv extends EventEmitter {
 		// Attach iterators
 		if (typeof this.opts.store[Symbol.iterator] === 'function' && this.opts.store instanceof Map) {
 			this.iterator = generateIterator(this.opts.store);
-		} else if (typeof this.opts.store.iterator === 'function') {
+		} else if (typeof this.opts.store.iterator === 'function' && this.opts.store.opts?.dialect === 'sqlite') { // Iterator only supported for sqlite for now
```
stropicall: @jaredwray added this since I only added for SQLITE for now. I wll add for remaining ones

packages/keyv/src/index.js
```diff
@@ -57,6 +57,9 @@ class Keyv extends EventEmitter {
 					? iterator(this.opts.store.namespace)
 					: iterator) {
 					const data = typeof raw === 'string' ? this.opts.deserialize(raw) : raw;
+					if (this.opts.store.namespace && !key.includes(this.opts.store.namespace)) {
+						continue;
```
stropicall, line 61: we are testing this line in `@Keyv/test-suite` and that is using `Map` as store but in `Keyv` project we are using SQLite store and we are handling same namespace keys in query so this line is not getting tested.

PR #238, opened 2022-02-09
implement iterables for postgres

### LINKED ISSUE #40: Expose Iterables

alphmth:
Fixes #40

### DIFF
--- a/packages/keyv/src/index.js
+++ b/packages/keyv/src/index.js
@@ -73,8 +73,8 @@ class Keyv extends EventEmitter {
 		// Attach iterators
 		if (typeof this.opts.store[Symbol.iterator] === 'function' && this.opts.store instanceof Map) {
 			this.iterator = generateIterator(this.opts.store);
-		} else if (typeof this.opts.store.iterator === 'function' && this.opts.store.opts && 
-					(this.opts.store.opts.dialect === 'sqlite' || this.opts.store.opts.dialect === 'postgres')) { // Iterator only supported for sqlite for now
+		} else if (typeof this.opts.store.iterator === 'function' && this.opts.store.opts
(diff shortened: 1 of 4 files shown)


jaredwray:
@alphmth - can you resolve the conflict?

packages/keyv/src/index.js
```diff
@@ -73,7 +73,8 @@ class Keyv extends EventEmitter {
 		// Attach iterators
 		if (typeof this.opts.store[Symbol.iterator] === 'function' && this.opts.store instanceof Map) {
 			this.iterator = generateIterator(this.opts.store);
-		} else if (typeof this.opts.store.iterator === 'function' && this.opts.store.opts && this.opts.store.opts.dialect === 'sqlite') { // Iterator only supported for sqlite for now
+		} else if (typeof this.opts.store.iterator === 'function' && this.opts.store.opts
```
jaredwray, line 82: @alphmth wonder if it would be better for right now to create an array for which ones support it above so you can just add a line each time such as the following:

```javascript
const iterableAdapters = [
'sqlite',
'postgres'
...
]

Then below you can just do a quick check if it is in the array.

jaredwray (changes requested):

jaredwray (approved):

PR #244, opened 2022-02-15
implement has()

### LINKED ISSUE #50: .has() api method
Am I missing something or there is no `.has()` api method? (similar to Map)

What I need is to distinguish between situation when there's no key in cache (`has` would return false) or there is a key in cache, but it's value is `undefined`. If I use `.get(key)` for this - both will return `undefined`.

alphmth:
Fixes #50

### DIFF
--- a/packages/keyv/src/index.js
+++ b/packages/keyv/src/index.js
@@ -162,6 +162,13 @@ class Keyv extends EventEmitter {
 		return Promise.resolve()
 			.then(() => store.clear());
 	}
+
+	has(key) {
+		const keyPrefixed = this._getKeyPrefix(key);
+		const { store } = this.opts;
+		return Promise.resolve()
+			.then(() => store.has(keyPrefixed));
+	}
 }
 
 module.exports = Keyv;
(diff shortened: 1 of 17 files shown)


jaredwray:
@alphmth - one of the things to look out for is if the storage adapter supports has. If it doesn't we should then try to do it via get with a `undefined` turning into false. If the function does exist then pass through the code. 

Remember that we have many 3rd party storage adapters that we support.

packages/keyv/src/index.js
```diff
@@ -162,6 +162,13 @@ class Keyv extends EventEmitter {
 		return Promise.resolve()
 			.then(() => store.clear());
 	}
+
+	has(key) {
```
jaredwray, line 166: @alphmth - what I was meaning here is what would happen if the storage adapter did not support the `has` function? Would it just error or would we handle it on our side?

Check if the function exists on the store, if not then use the get function and return true or false, if it does exist use it.

stropicall: @jaredwray all storage adapters in this repo support `has`

Should I still handle the case?

stropicall: @jaredwray I have supported that case now I think. Please check

PR #254, opened 2022-03-01
get many

### LINKED ISSUE #78: Get multiple keys in a single operation
Similar to https://github.com/lukechilds/keyv/issues/41 but with `.get()` instead of `.delete()`

Pass an array of keys to get in a single operation.

```js
await keyv.get(['key-1', 'key-2', 'key-3']);
```

alphmth:
Fixes #78

### DIFF
--- a/packages/etcd/src/index.js
+++ b/packages/etcd/src/index.js
@@ -65,8 +65,11 @@ class KeyvEtcd extends EventEmitter {
 			.then(values => {
 				const data = [];
 				for (const value of values) {
-				    if (value.value === null) data.push(undefined);
-				    else data.push(value.value);
+					if (value.value === null) {
+						data.push(undefined);
+					} else {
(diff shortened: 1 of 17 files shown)


trevor-scheer:
Hey @alphmth and @jaredwray - I'm digging into a bug that we're running into over here:
https://github.com/apollographql/apollo-utils/issues/166

I'm fairly confident at this point that this isn't `ioredis`'s fault and that the `getMany` implementations introduced here are the problem - returning an empty array in the all-cache-miss case is problematic for us. I'm curious if there was a particular reason it was decided to be that way? It seems like this incurs additional overhead anyway since every `getMany` call has to iterate over the full array to check if it's all cache misses. For our case, an array of `null`s would be preferable.

Would you accept a PR that changes this behavior to remove the special all-cache-miss case?

PR #267, opened 2022-04-02
fix typescript issues (old node + previous @types/keyv compat issues)

### LINKED ISSUE #268: difference between @types/keyv and types of keyv
With https://github.com/jaredwray/keyv/pull/257 typescript support was implemented. I love it and support it. However, this was a breaking change in my eyes. The types differ between [@types/keyv](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/6963e48c635515344d0f1259cfdecb163390cf4e/types/keyv/index.d.ts) and [keyv](https://github.com/jaredwray/keyv/blob/main/packages/keyv/src/index.d.ts) package. 
I use `keyv` myself only as transitive dependency via the `cachable-request`. This imports keyv via [`>4.0.0`](https://github.com/lukechilds/cacheable-request/blob/master/package.json#L35). After `npm upgrade` I get this error.

```
node_modules/@types/cacheable-request/index.d.ts:13:10 - error TS2305: Module '"keyv"' has no exported member 'Store'.

13 import { Store } from 'keyv';
```

typescript prioritize the shipped types over types provided with ' @types/...'.  Would it be possible to add the missing types, or is this not desired (could provide pr)? Another solution would be to update to a new major version and revert the change in v4.x.x

d-fischer:
Hello, I'm currently having problems using a dependent of this. The cause is that imports are being made from `node:events`, a notation that is only supported in very new versions of the `@types/node` package (I think 16.6 and up). To support node 12 and 14, I changed the imports to `events` and used the named import (the default import was actually broken for me).

Additionally, I resolved a compatibility issue for people that used the `@types/keyv` package - the new definitions were missing the `Store` type. Aliasing it to the "new" `Keyv` type works fine for now, but there should be some minor refactoring for this in the future, probably.

Fixes #268

### DIFF
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -1,9 +1,9 @@
-import EventEmitter from 'events';
+import { EventEmitter } from 'events';
 
-declare class Keyv extends EventEmitter {
+declare class Keyv<T = any> extends EventEmitter {
 	constructor(uri: string, options: Keyv.Options);
-	get(key: string, options: any): Promise<any>;
-	set(key: string, value: any, ttl: number): Promise<boolean>;
+	get(key: string, options: any): Promise<T>;
+	set(key: string, value: T, ttl: number): Promise<boolean>;
 	delete(key: string): Promise<boolean>;
 	clear(): Promise<void>;
 	has(key: string): Promise<boolean>;
@@ -16,6 +16,8 @@ declare namespace Keyv {
 		deserialize?: ((data: string) => any | undefined) | undefined;
 		ttl?: number | undefined;
 	}
+
+	type Store<T> = Keyv<T>;
 }
 
 export = Keyv;
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -2,8 +2,8 @@ import { EventEmitter } from 'events';
 
 declare class Keyv<T = any> extends EventEmitter {
 	constructor(uri: string, options: Keyv.Options);
-	get(key: string, options: any): Promise<T>;
-	set(key: string, value: T, ttl: number): Promise<boolean>;
+	get(key: string, options?: any): Promise<T>;
+	set(key: string, value: T, ttl?: number): Promise<boolean>;
 	delete(key: string): Promise<boolean>;
 	clear(): Promise<void>;
 	has(key: string): Promise<boolean>;
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -1,4 +1,4 @@
-import EventEmitter from 'node:events';
+import EventEmitter from 'events';
 
 declare class Keyv extends EventEmitter {
 	constructor(uri: string, options: Keyv.Options);
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -1,4 +1,4 @@
-import { EventEmitter } from 'events';
+import {EventEmitter} from 'events';
 
 declare class Keyv<T = any> extends EventEmitter {
 	constructor(uri: string, options: Keyv.Options);
--- a/package.json
+++ b/package.json
@@ -19,5 +19,8 @@
   },
   "dependencies": {
     "@types/node": "^17.0.21"
+  },
+  "devDependencies": {
+    "lerna": "^4.0.0"
   }
 }
--- a/packages/etcd/package.json
+++ b/packages/etcd/package.json
@@ -13,7 +13,8 @@
 		"rules": {
 			"unicorn/prefer-module": 0,
 			"unicorn/no-array-reduce": 0,
-			"unicorn/prefer-object-from-entries": 0
+			"unicorn/prefer-object-from-entries": 0,
+			"unicorn/prefer-node-protocol": 0
 		}
 	},
 	"ava": {
--- a/packages/etcd/package.json
+++ b/packages/etcd/package.json
@@ -9,7 +9,6 @@
 		"clean": "rm -rf node_modules && rm -rf .nyc_output && rm -rf coverage.lcov"
 	},
 	"xo": {
-		"extends": "xo-lukechilds",
 		"rules": {
 			"unicorn/prefer-module": 0,
 			"unicorn/no-array-reduce": 0,
@@ -51,7 +50,6 @@
 	"devDependencies": {
 		"@keyv/test-suite": "*",
 		"ava": "^4.0.1",
-		"eslint-config-xo-lukechilds": "^1.0.1",
 		"keyv": "*",
 		"nyc": "^15.1.0",
 		"requirable": "^1.0.5",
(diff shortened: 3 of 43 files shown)


packages/keyv/src/index.d.ts
```diff
@@ -1,9 +1,9 @@
-import EventEmitter from 'node:events';
+import { EventEmitter } from 'events';
 
-declare class Keyv extends EventEmitter {
+declare class Keyv<T = any> extends EventEmitter {
 	constructor(uri: string, options: Keyv.Options);
-	get(key: string, options: any): Promise<any>;
-	set(key: string, value: any, ttl: number): Promise<boolean>;
+	get(key: string, options: any): Promise<T>;
```
jmopel:
```suggestion
	get(key: string, options?: any): Promise<T>;
```

packages/keyv/src/index.d.ts
```diff
@@ -1,9 +1,9 @@
...
+	set(key: string, value: T, ttl: number): Promise<boolean>;
```
jmopel:
```suggestion
	set(key: string, value: T, ttl?: number): Promise<boolean>;
```

d-fischer:
It is literally impossible to fix some of these errors without touching the linter config. You have conflicting rules set up. `@typescript-eslint/object-curly-spacing` is set up to prohibit spacing, while `object-curly-spacing` requires them.

So, how do you actually want it?

d-fischer:
I see that the "override config" you're extending literally contains only `object-curly-spacing`, which conflicts with the TypeScript defaults. I'll see what I can do to fix this. I'll probably remove that base config.

d-fischer:
Apparently, `xo` does something weird with the rules. For TS files only, it loads a separate ruleset and additional plugins. So I can't add the `@typescript-eslint/*` rule, because it would break all non-TS files. 

There's a few different ways to fix this:

- Add a `tsconfig.json` file to every package (which probably should be the case anyway) and enable the `ts` flag in the `xo` config
- Opt for the default setting (no spaces allowed inside curly braces)
- Get rid of `xo` and set up the linter manually (my personal favorite, but probably unlikely to happen 😉)

d-fischer:
Okay, option 1 is out of the question, since there's a bunch of idioms in the JS files that just fail to check against the default TypeScript rules... Going for option 2 then.

d-fischer:
Okay, this has become much more invasive than I thought it would. Hope it's still to your liking and I can go on with my own projects soon.

d-fischer:
If you want, I can undo the last commit, and you can then go and override the actions results, it's up to you.

pavelkalin:
Guys, 

Thank you for working on this bug. I've no idea which package is relying on keyv.

I also now face critical issue with:
node_modules/@types/cacheable-request/index.d.ts(13,10): error TS2305: Module '"keyv"' has no exported member 'Store'.
node_modules/keyv/src/index.d.ts(1,8): error TS1259: Module '"node:events"' can only be default-imported using the 'esModuleInterop' flag

I'll post it here for whoever else has it and maybe will find this thread in future. My esModuleInterop is true

jaredwray:
> 

@d-fischer first, I want to thank you so much for taking the lead on this and starting to get a fix going for the types issue showing an error: 

```
node_modules/@types/cacheable-request/index.d.ts(13,10): error TS2305: Module '"keyv"' has no exported member 'Store'.
node_modules/keyv/src/index.d.ts(1,8): error TS1259: Module '"node:events"' can only be default-imported using the 'esModuleInterop' flag
```

I would like to roll back all the changes except setting the type definition locally and fixing it to handle Store. Can we just fix that in the index.d.ts file? That is the major change on the 4.2.0 release is that we are now having local type definition files. 

Once that is completed and fixed we can move to the next update which is if we should remove node:events to just events and run that as another pull request. 

@alphmth - looks like your latest pull request with moving type definition locally is causing an error which we need to resolve ASAP. 

@pavelkalin - go ahead and stay on the previous version until we resolve this.

d-fischer:
`node:events` is not available if you use an older `@types/node` package. The one you are using is irrelevant, TS prioritizes the one that's in the end user's package.json. This is a breaking change and should be fixed ASAP, it shouldn't be reverted. I can revert the linter changes that make the Actions fail though.

d-fischer:
Also, it's hard to just "stay on the previous version". For me, this is a deeply nested dependency and I have no control over its version range. This blocks me.

d-fischer:
For the record, this is the console output if you have `@types/node@^12` installed:

```
../../node_modules/@types/cacheable-request/index.d.ts:13:10 - error TS2305: Module '"keyv"' has no exported member 'Store'.
../../node_modules/keyv/src/index.d.ts:1:26 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.
```

jaredwray:
> For the record, this is the console output if you have `@types/node@^12` installed, with the version of `keyv` that's currently on npm:
> 
> ```
> ../../node_modules/@types/cacheable-request/index.d.ts:13:10 - error TS2305: Module '"keyv"' has no exported member 'Store'.
> ../../node_modules/keyv/src/index.d.ts:1:26 - error TS2307: Cannot find module 'node:events' or its corresponding type declarations.
> ```

@d-fischer Running the checks now. Do you believe this pull request is ready to be tested against?

d-fischer:
I already did all the checks locally. They fail with a lot of linter errors, which is what the last commit was about, which you insisted I revert.

Before reverting, all the checks were fine on my local system, including both linter and tests.

jaredwray:
> I already did all the checks locally. They fail with a lot of linter errors, which is what the last commit was about, which you insisted I revert.
> 
> Before reverting, all the checks were fine on my local system, including both linter and tests.

OK. please apply all changes back in so I can validate. Sorry for the hassle on this as I wanted to get this in a better separation.

jaredwray:
@d-fischer - running checks and validating code now.

jaredwray:
@d-fischer - are you also using `cacheable-request`?

d-fischer:
> @d-fischer - are you also using `cacheable-request`?

My dependency chain according to yarn:

```
ngrok#got#cacheable-request#keyv
```

adityap31:
I added, "keyv": "4.1.1"  as a dependency to make it work. It works now. Workaround definitely.

AnWeber:
@adityap31 The release of 4.2.1 allowed me to remove my workarounds.

pavelkalin:
Yes, thank you! no more issues from my side as well.

packages/keyv/src/index.d.ts
```diff
@@ -16,6 +16,8 @@ declare namespace Keyv {
 		deserialize?: ((data: string) => any | undefined) | undefined;
 		ttl?: number | undefined;
 	}
+
+	type Store<T> = Keyv<T>;
```
trevor-scheer, line 20: Is this accurate? In the existing types `Store` was a simpler interface. Does it actually need to be a `Keyv`?

PR #272, opened 2022-04-06
Add valid constructor calls and arguments to types

### LINKED ISSUE #271: KeyV constructor type does not fully express its valid uses
Got a dependency bump PR for keyv over here: https://github.com/backstage/backstage/pull/10635 and it's failing on typescript errors for valid constructor calls, e.g.

```
25     client = new Keyv();
                ~~~~~~~~~~

  node_modules/keyv/src/index.d.ts:4:14
    4  constructor(uri: string, options: Keyv.Options);
                   ~~~~~~~~~~~
    An argument for 'uri' was not provided.
```

```
133     return new Keyv({
               ~~~~~~~~~~
134       namespace: pluginId,
    ~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
136       store: new KeyvRedis(this.connection),
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
137     });
    ~~~~~~

  node_modules/keyv/src/index.d.ts:4:27
    4  constructor(uri: string, options: Keyv.Options);
                                ~~~~~~~~~~~~~~~~~~~~~
    An argument for 'options' was not provided.
```

iamEAP:
~~This covers my use-cases and most (all?) of those enumerated in the `README.md`, but doesn't necessarily cover everything the constructor actually may handle (e.g. compression options).~~

This copies over types from [https://github.com/DefinitelyTyped/DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/keyv/index.d.ts)

Closes #271

### DIFF
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -1,23 +1,82 @@
 import {EventEmitter} from 'events';
 
-declare class Keyv<T = any> extends EventEmitter {
-	constructor(uri: string, options: Keyv.Options);
-	get(key: string, options?: any): Promise<T>;
-	set(key: string, value: T, ttl?: number): Promise<boolean>;
+type WithRequiredProperties<T, K extends keyof T> = T & Required<Pick<T, K>>;
+
+declare class Keyv<Value = any, Options extends Record<string, any> = Record<string, unknown>> extends EventEmitter {
+	/**
+     * `this.opts` is an object containing at least the properties listed
+     * below. However, `Keyv.Options` allows arbitrary properties as well.
+     * These properties can be specified as the second type parameter to `Keyv`.
+     */
+	opts: WithRequiredProperties<
+	Keyv.Options<Value>,
+	'deserialize' | 'namespace' | 'serialize' | 'store' | 'uri'
+	> &
+	Options;
+
+	/**
+     * @param opts The options object is also passed through to the storage adapter. Check your storage adapter docs for any extra options.
+     */
+	constructor(options?: Keyv.Options<Value> & Options);
+	/**
+     * @param uri The connection string URI.
+     *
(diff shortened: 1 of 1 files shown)


thibauds:
Maybe we could use https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/keyv/index.d.ts ?

jaredwray:
> https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/keyv/index.d.ts

I believe we should just move to having this file in the local repository. Do you want to move to that in this pull request and have us all validate it? 

The goal is to move more of this into the same repository and eventually we are considering moving to typescript but that will take time and we are weighing the benefits.

iamEAP:
> Do you want to move to that in this pull request and have us all validate it?

Sounds good.  Done!  Please take a look when you've got a chance.

jaredwray:
> > Do you want to move to that in this pull request and have us all validate it?
> 
> Sounds good. Done! Please take a look when you've got a chance.

I think you will need to run ‘xo —fix’ on this file

iamEAP:
👍  `xo` was complaining on the naming of the generics (not strictly pascal case'd), so I chose to just remove the `T` prefixes.  Descriptive naming seems to be the recommendation: https://github.com/xojs/eslint-config-xo-typescript/issues/41

jaredwray:
@iamEAP can you validate this change fixes the issue?

trevor-scheer:
🤦‍♂️ sorry for the ping, I only looked at the GH releases and assumed this wasn't released. Trying out 4.2.2 now.

trevor-scheer:
@jaredwray lgtm, thanks!

PR #275, opened 2022-04-08
covered all lines

### LINKED ISSUE #274: Testing for the Keyv package on the following lines 70,107,134,179-183
----------|---------|----------|---------|---------|--------------------
File      | % Stmts | % Branch | % Funcs | % Lines | Uncovered Line #s  
----------|---------|----------|---------|---------|--------------------
All files |     100 |    93.69 |     100 |     100 |                    
 index.js |     100 |    93.69 |     100 |     100 | 70,107,134,179-183 
----------|---------|----------|---------|---------|--------------------

alphmth:
Fixes #274

### DIFF
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -27,7 +27,7 @@ declare class Keyv<Value = any, Options extends Record<string, any> = Record<str
 	constructor(uri?: string, options?: Keyv.Options<Value> & Options);
 
 	/** Returns the value. */
-	get<Raw extends boolean = false>(key: string, options?: {raw?: Raw}):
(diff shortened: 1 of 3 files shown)


glasser:
@jaredwray This is a backwards-incompatible change: custom Stores now need to provide `has`, which wasn't the case before. Seems like this should have had a major version bump? It breaks our code which has a custom Store without `has`.

jaredwray:
@glasser - so sorry to hear that this is a breaking change. I thought we made it so that if has was not detected we handled it a different way. Do you have an example of how it is breaking so that I can get a fix going?

glasser:
It's breaking for Typescript users because it adds a mandatory method to
Stores that you pass in. If it's not actually mandatory you can put a ?
after the method name.

PR #283, opened 2022-04-12
Bug 278

### LINKED ISSUE #278: Redis constructor type does not fully express its valid uses
[`@keyv/redis`'s constructor signature](https://github.com/jaredwray/keyv/blob/main/packages/redis/src/index.d.ts#L7) does not support passing it a ` Redis.Cluster` or `Redis` instance which looks like it's supported based on the [implementation](https://github.com/jaredwray/keyv/blob/main/packages/redis/src/index.js#L11)

`@keyv/redis` version `2.3.1`

alphmth:
Fixes #278


jaredwray:
@alphmth - can you resolve the conflicts?

Issue #256, opened 2022-03-08
Single quote in value results in SQLITE_ERROR


dshepsis:
### Problem
When using `Keyv#set`, an error is thrown if the `value` contains a string containing a single quote.

For example, if the string is simply a single quote, the error is:
```
node:internal/process/promises:246
          triggerUncaughtException(err, true /* fromPromise */);
          ^

[Error: SQLITE_ERROR: near "","": syntax error] {
  errno: 1,
  code: 'SQLITE_ERROR'
}
```

### How to reproduce
Run this demonstration code in a directory with Keyv added as a dependency:
```js
const Keyv = require('keyv');

const testDB = new Keyv('sqlite://test.sqlite', { namespace: 'test' });
testDB.on('error', err => console.log('Connection Error', err));

const value = "'";
testDB.set('key', value);
```
This will result in the above error.

It seems that https://github.com/jaredwray/keyv/pull/210 was supposed to address this issue, but it's either not part of the current release, or failed to fix the issue. Either way, a test case should probably be added for this.

### Expected behavior
Values containing single quotes in strings should be serialized normally, like any other values.

### Version
4.1.1

alphmth:
@dshepsis can you pas it like this`"\'"` or `'\''`

dshepsis:
> @dshepsis can you pas it like this`"\'"` or `'\''`

This doesn't work for a few reasons. Of course, just using a single backslash has no effect because the single quote is already a literal in this case. In other words, the following code produces exactly the same result as the code in the original report:

```js
const Keyv = require('keyv');

const testDB = new Keyv('sqlite://test.sqlite', { namespace: 'test' });
testDB.on('error', err => console.log('Connection Error', err));

const value = "\'";
testDB.set('key', value);
```

If you instead use a double backslash `"\\'"`, it produces the same error yet again (not really sure why. Maybe backslash isn't a valid escape character in this SQL implementation?). Using a double single quote `"''"` does actually work, with the saved result having just a single single quote, since a double single quote is a valid escape sequence for a single single quote in this case.

This isn't an acceptable workaround either, because I'm trying to store objects instead of strings. To escape everything would require iterating through every single key and value of the object, recursively, to double-up every single quote. Also, this means that anyone trying to pass in strings/objects which already contain double single quotes (for whatever reason) are having their data silently modified by keyv.

It's also worth mentioning that the key passed into the `.set` method also can't contain single quotes:
```js
const Keyv = require('keyv');

[...]

Issue #221, opened 2022-01-30
From the last update of @keyv/sqlite, suddenly its unable to work properly.


ShreshthTiwari:
Connection Error [Error: SQLITE_ERROR: near "907626718095896626": syntax error] {
  errno: 1,
  code: 'SQLITE_ERROR'
}
/home/shreshthtiwari/Documents/GitHub/B0T/node_modules/@keyv/sqlite/src/index.js:44
                        .then(query => query(sqlString));
                                       ^

TypeError: query is not a function
    at /home/shreshthtiwari/Documents/GitHub/B0T/node_modules/@keyv/sqlite/src/index.js:44:19
    at async Client.<anonymous> (/home/shreshthtiwari/Documents/GitHub/B0T/index.js:176:27)

Node.js v17.3.0

jaredwray:
@ShreshthTiwari is this right on when loading node modules or when you are calling it? If so, can you provide any code as reference?

jaredwray:
@ShreshthTiwari one thought on this. Have you validated it with node v16.x.x? That is what we are testing towards at this time.

ShreshthTiwari:
I'm using the latest version 17 maybe.
The code is here 
https://github.com/ShreshthTiwari/B0T/blob/main/builders/databaseBuilder.js

ShreshthTiwari:
And th error is this-
![Screenshot_2022_0131_233722](https://user-images.githubusercontent.com/68736752/151848756-851705ac-d871-476a-b9ba-b092a8bdd572.jpg)

jaredwray:
Are you using NVM (https://github.com/nvm-sh/nvm)? If so, please move it back to v16 and try there. We are looking into this also.

ShreshthTiwari:
Tried in v16 😔
![Uploading Screenshot_2022_0131_234111.jpg…]()

ShreshthTiwari:
![Screenshot_2022_0131_234111](https://user-images.githubusercontent.com/68736752/151849420-53b4999a-1b68-47b6-a0b0-21a5f5ee9d6e.jpg)

alphmth:
@jaredwray we are now using `better-sqlite3`. Was that released or it has been released yet?

jaredwray:
@alphmth - it has been released. can you work to get it working correctly and add in tests to make sure it works?

jaredwray:
[...]

PR #290, opened 2022-04-18
change better sqlite to sqlite3

### LINKED ISSUE #286: Last version to use sqlite3
What is the last version to support sqlite3? better-sqlite3 does not support `<db>.opts.store.query` and has broken large portions of my projects due to this. 

Trying to find out the last version(s) of keyv and keyv/sqlite that used sqlite3 rather than better-sqlite3.

Related to #221

alphmth:
Fixes #286 #256 #221


jaredwray:
@alphmth - can you also look at building unit tests and validation for the other issues: #256 and #221 

Please add additional unit tests that validate these scenarios are working properly.

alphmth:
@jaredwray can you add `--concurrency 1` to the `test` script of the root package and check?

jaredwray:
> @jaredwray can you add `--concurrency 1` to the `test` script of the root package and check?

`--concurrency=1` is what I did and that fixed it.

PR #307, opened 2022-04-29
close connection for redis

alphmth:


jaredwray:
@alphmth - this looks good. One thought around this is if we should make them all `async` as it might be needed by other 3rd party libraries and drivers.

Also, it would be good to update the readme with this update on the Keyv package with `async disconnect()`

alphmth:
@jaredwray all drivers `disconnect` method returns promise already.

I will update readme

PR #328, opened 2022-05-10
Add disconnect to type declarations

gafderks:
Related: #307


jaredwray:
@gafderks looks like there is a spacing issue. Can you resolve?

PR #330, opened 2022-05-12
optional to support backward compatibilty

alphmth:


jaredwray:
@glasser - would this do the fix you need?

glasser:
This looks good!

glasser:
disconnect is a function that you define on keyv. You can add new functionality and that's fine; we may or may not use it.

`has` is a method on `Store`, which is something *we* make and *provide* to keyv. Adding new required methods to `Store` is thus backwards-compatible.

Put it this way: let's say you added this method to the `Store` interface: `jklsdafadsfhjkasdfasdf(): void`.  It seems very unlikely that anyone has ever written any TypeScript class that defines this method.  So no existing code will be defining a class (or other type) that implements the new version of the `Store` method. That means no existing code calling `new Keyv({ store })` will properly typecheck after upgrading.

The line you linked is in the JavaScript code, not the type definitions. You're right that *at runtime* there wouldn't be problems. The problems my code ran into were at TypeScript type-checking time.

jaredwray:
@glasser - going to push that live.

trevor-scheer:
@jaredwray This doesn't seem to have taken effect in the most recent patch release, though I'm not sure why. Ddi you release from the wrong branch maybe?

https://unpkg.com/keyv@4.2.8/src/index.d.ts

glasser:
Agreed, this fix doesn't seem to have been published.

jaredwray:
@glasser - the fix is now live with `4.2.9` https://unpkg.com/keyv@4.2.9/src/index.d.ts

PR #349, opened 2022-05-27
Added async deserialize/serialize function types

cesarfd:
Added `Promise`d return types for `serialize` and `deserialize`.

### DIFF
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -65,8 +65,8 @@ declare namespace Keyv {
 		namespace?: string | undefined;
 		/** A custom serialization function. */
 		serialize?: ((data: DeserializedData<Value>) => string | Promise<string>) | undefined;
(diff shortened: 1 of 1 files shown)


kaioduarte:
@jaredwray FYI, this change is breaking the Apollo server adapter

```ts
The types returned by 'serialize(...)' are incompatible between these types.
    Type 'string | Promise<string>' is not assignable to type 'string'.
        Type 'Promise<string>' is not assignable to type 'string'.
```

jaredwray:
@kaioduarte - thanks for bringing this to our attention. Do you know what the recommended fix is? 

If not, I can dig in later this week to figure out what is the difference.

jaredwray:
@kaioduarte - we have a fix that should go out for `4.3.2` today we believe. It is now on `main` if you want to validate.

kaioduarte:
@jaredwray amazing, thank you!

PR #359, opened 2022-06-13
Add multi-key get typings overload

trevor-scheer:
`Keyv` supports the use of `get` in an overloaded manner. Pass a single key, get a single result; pass an array of keys, get an array of results.

The current typings as they're written return a singular `Value` in both cases. By overloading this method, we can get the correct return type `string -> Value | undefined`, `string[] -> (Value | undefined)[]`.

### DIFF
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -27,10 +27,19 @@ declare class Keyv<Value = any, Options extends Record<string, any> = Record<str
 	constructor(uri?: string, options?: Keyv.Options<Value> & Options);
 
 	/** Returns the value. */
-	get<Raw extends boolean = false>(key: string | string[], options?: {raw?: Raw}):
+	get<Raw extends boolean = false>(key: string, options?: {raw?: Raw}):
 	Promise<(Raw extends false
 		? Value
 		: Keyv.DeserializedData<Value>) | undefined>;
+
+	/** getMany by providing an array of keys */
+	get<Raw extends boolean = false>(
+		key: string[],
+		options?: { raw?: Raw }
+	): Promise<
+		((Raw extends false ? Value : Keyv.DeserializedData<Value>) | undefined)[]
+	>;
+	
 	/**
      * Set a value.
      *


jaredwray:
@trevor-scheer - can you get the linting fixed on these? Most likely you can do a `xo --fix`

trevor-scheer:
@jaredwray it did fix some things but also spit out these errors (unrelated to my changes)
```
trevorscheer@Trevors-MBP keyv % npx xo --fix

  packages/memcache/test/test.js:1:21
  ✖    1:21  Do not use "require".                                                           unicorn/prefer-module
  ✖    2:14  Do not use "require".                                                           unicorn/prefer-module
  ✖    3:14  Do not use "require".                                                           unicorn/prefer-module
  ✖    4:22  Do not use "require".                                                           unicorn/prefer-module
  ✖    6:14  Do not use "require".                                                           unicorn/prefer-module
  ✖   12:1   Do not use "require".                                                           unicorn/prefer-module
  ✖   16:5   Unexpected use of the global variable process. Use require("process") instead.  node/prefer-global/process
  ✖   17:8   Unexpected use of the global variable process. Use require("process") instead.  node/prefer-global/process
  ✖   32:33  Function was declared as async but doesn't use await.                           ava/no-async-fn-without-await

  packages/memcache/src/index.js:1:1
  ✖    1:1   Do not use "use strict" directive.                                              unicorn/prefer-module
  ✖    3:22  Do not use "require".                                                           unicorn/prefer-module
  ✖    4:18  Do not use "require".                                                           unicorn/prefer-module
  ✖    5:15  Do not use "require".                                                           unicorn/prefer-module
  ✖  143:32  reject is defined but never used. Allowed unused args must match /^_/u.         no-unused-vars
  ✖  155:1   Do not use "module".                                                            unicorn/prefer-module

  packages/memcache/src/index.d.ts:13:26
  ✖   13:26  any overrides all other types in this union type.                               @typescript-eslint/no-redundant-type-constituents

  packages/mongo/src/index.js:134:43
  ✖  134:43  Prefer utf8 over utf-8.                                                         unicorn/text-encoding-identifier-case

  17 errors
```

jaredwray:
@trevor-scheer - thanks go ahead and roll back the changes and we can fix them on our side then will bring in your type fix.

jaredwray:
Looks like xo version `0.50.0` has some breaking changes that needs to fix and also `memcache` package needs to get updated to be compliant. 

to test the project just download the latest and then do the following:

1. run `yarn`
2. run `yarn test:services:start` make sure you have docker installed and running
3. Wait about 30 seconds for all the services to start up
4. run `yarn test` -- this will test all the major packages with the current setup including `xo` lint

jaredwray:
@trevor-scheer - this will be released in the next week or so with other updates being tested

trevor-scheer:
Thanks!

PR #362, opened 2022-06-14
Add missing `getMany` optional method on Store interface

trevor-scheer:
I'm a little unsure about the `StoredData` type I landed on, it's what I gathered based on the src here:
https://github.com/jaredwray/keyv/blob/main/packages/keyv/src/index.js#L155-L174

Also, `xo` is breaking my type:
I want:
`type StoredData<Value> = DeserializedData<Value> | string | null | undefined;`
...but it mistakenly replaces with:
`type StoredData<Value> = DeserializedData<Value> | string | undefined | undefined;`

Just a consideration: `prettier` tends to be the formatting standard in JS/TS and I've never seen an issue as glaring as ^ that while using it.

### DIFF
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -92,12 +92,17 @@ declare namespace Keyv {
 		value: Value; expires: number | undefined;
 	}
 
+	type StoredData<Value> = DeserializedData<Value> | string | undefined | undefined;
+
 	interface Store<Value> {
 		get(key: string): Value | Promise<Value | undefined> | undefined;
 		set(key: string, value: Value, ttl?: number): any;
 		delete(key: string): boolean | Promise<boolean>;
 		clear(): void | Promise<void>;
 		has?(key: string): boolean | Promise<boolean>;
+		getMany?(
+			keys: string[]
+		): Array<StoredData<Value>> | Promise<Array<StoredData<Value>>> | undefined;
 	}
 }
 


packages/keyv/src/index.d.ts
```diff
@@ -92,12 +92,17 @@ declare namespace Keyv {
 		value: Value; expires: number | undefined;
 	}
 
+	type StoredData<Value> = DeserializedData<Value> | string | undefined | undefined;
```
glasser, line 95: Why two undefineds?

glasser: oh geez i see. whoa.

jaredwray:
> I'm a little unsure about the `StoredData` type I landed on, it's what I gathered based on the src here: https://github.com/jaredwray/keyv/blob/main/packages/keyv/src/index.js#L155-L174
> 
> Also, `xo` is breaking my type: I want: `type StoredData<Value> = DeserializedData<Value> | string | null | undefined;` ...but it mistakenly replaces with: `type StoredData<Value> = DeserializedData<Value> | string | undefined | undefined;`
> 
> Just a consideration: `prettier` tends to be the formatting standard in JS/TS and I've never seen an issue as glaring as ^ that while using it.

@trevor-scheer thanks for the advice and this project started with xo so we have stuck with it. Agree that prettier might be better but many of us are used to `xo`.

trevor-scheer:
@jaredwray i don't think this was ready to merge. My comment mentioned the double `undefined | undefined` problem when it should've been `undefined | null`. That needs to be fixed in a follow up or the original commit reverted.

jaredwray:
@trevor-scheer - reverting this now. Thanks for the heads up as I picked the wrong one to merge.

jaredwray:
@trevor-scheer go ahead and open a new pull request. Again, sorry for the hassle.

PR #384, opened 2022-06-22
Reintroduce `getMany` optional method

trevor-scheer:
Reinstates https://github.com/jaredwray/keyv/pull/362

The `undefined | undefined` type was happening via `xo`'s fix option due to a lint config which forbids the use of `null` (and converts to `undefined` (`@typescript-eslint/ban-types`). In hindsight, I'm not sure the `null` was correct there to begin with.


jaredwray:
@trevor-scheer - it looks like this definition was rolled back because it broke other clients.

Issue #404, opened 2022-07-12
Keyv.Options type incorrectly defines store option type


dylanseago:
**Describe the bug**
The `Keyv.Options` type requires a type of `Store<Value>`, but all stores in this library appear to be implementing `Store<string | undefined>`. If the `Value` type is anything but `string` then using this option isn't possible, typescript will complain. This should be expecting a `Store<string | undefined>`, which also makes sense considering all values are, by default, serialized to string using JSONB before being passed down to the store.

**To Reproduce**
Via repo:
```
git clone git@github.com:dylanseago/keyv.git
git checkout repro-bad-store-types
yarn test:services:start
yarn test
```

Error:
```
  Uncaught exception in test/types.ts

  test/types.ts(13,15): error TS2769: No overload matches this call.
    Overload 1 of 2, '(options?: (Options<MyType> & Record<string, unknown>) | undefined): Keyv<MyType, Record<string, unknown>>', gave the following error.
      Type 'KeyvRedis' is not assignable to type 'Store<string | undefined>'.
        The types returned by 'getMany(...)' are incompatible between these types.
          Type 'Promise<string[] | undefined>' is not assignable to type 'StoredData<string | undefined>[] | Promise<StoredData<string | undefined>[]> | undefined'.
            Type 'Promise<string[] | undefined>' is not assignable to type 'Promise<StoredData<string | undefined>[]>'.
              Type 'string[] | undefined' is not assignable to type 'StoredData<string | undefined>[]'.
                Type 'undefined' is not assignable to type 'StoredData<string | undefined>[]'.
    Overload 2 of 2, '(uri?: string | undefined, options?: (Options<MyType> & Record<string, unknown>) | undefined): Keyv<MyType, Record<string, unknown>>', gave the following error.
      Argument of type '{ store: KeyvRedis; }' is not assignable to parameter of type 'string'.


  ✖ test/types.ts exited with a non-zero exit code: 1
  ─

  49 tests passed
  1 uncaught exception
```

**Expected behavior**
You should be able to instantiate a `Keyv` instance using `KeyvRedis` store in typescript.

**Tests (Bonus!!!)**
See https://github.com/dylanseago/keyv/blob/repro-bad-store-types/packages/redis/test/types.ts

**Additional context**
This bug likely impacts using more than just `KeyvRedis`, but that's what I'm trying to get working in typescript right now so it's what I added a test for.

iamEAP:
[...]

PR #405, opened 2022-07-13
Fix Keyv.Options type incorrectly defines store option type

dylanseago:
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/jaredwray/keyv/blob/main/CODE_OF_CONDUCT.md

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fix https://github.com/jaredwray/keyv/issues/404

* **What is the current behavior?** (You can also link to an open issue here)
https://github.com/jaredwray/keyv/issues/404

* **What is the new behavior (if this is a feature change)?**

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
This may impact some typescript users 

* **Other information**:

### DIFF
--- a/packages/redis/src/index.d.ts
+++ b/packages/redis/src/index.d.ts
@@ -10,7 +10,7 @@ declare class KeyvRedis extends EventEmitter implements Store<string | undefined
 	constructor(options?: KeyvRedis.Options | Redis | Cluster);
 	constructor(uri: string | Redis | Cluster, options?: KeyvRedis.Options);
 	get(key: string): Promise<string | undefined>;
-	getMany(keys: string[]): Promise<string[] | undefined>;
+	getMany(keys: string[]): Promise<string[]>;
 	set(key: string, value: string | undefined, ttl?: number): Promise<any>;
 	delete(key: string): boolean;
 	deleteMany(keys: string[]): boolean;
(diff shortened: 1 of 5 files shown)


packages/redis/src/index.d.ts
```diff
@@ -10,7 +10,7 @@ declare class KeyvRedis extends EventEmitter implements Store<string | undefined
 	constructor(options?: KeyvRedis.Options | Redis | Cluster);
 	constructor(uri: string | Redis | Cluster, options?: KeyvRedis.Options);
 	get(key: string): Promise<string | undefined>;
-	getMany(keys: string[]): Promise<string[] | undefined>;
+	getMany(keys: string[]): Promise<string[]>;
```
jaredwray, line 13: @dylanseago - have you validated this is correct as I believe that in some cases it can return an undefined.

dylanseago: @jaredwray yes, as per the implementation, it uses `Array.every` on the `rows` returned from redis. So if it were possible for `undefined` to be returned here then this would be throwing a `TypeError` when trying to call on `undefined`.
https://github.com/jaredwray/keyv/blob/main/packages/redis/src/index.js#L38

Although, it appears to be checking if any individual _values_ are `null`. So it may be more correct to return `Promise<Array<string | null>>` instead

PR #417, opened 2022-07-19
Issue 403

### LINKED ISSUE #404: Keyv.Options type incorrectly defines store option type

alphmth:
Fixes #404 
* **Please check if the PR fulfills these requirements**
- [x] The commit message follows our guidelines: https://github.com/jaredwray/keyv/blob/main/CODE_OF_CONDUCT.md

* **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
* Bug fix

* **What is the current behavior?** (You can also link to an open issue here)
* The Keyv.Options type requires a type of Store<Value>, but all stores in this library appear to be implementing Store<string | undefined>. If the Value type is anything but string then using this option isn't possible, typescript will complain. This should be expecting a Store<string | undefined>, which also makes sense considering all values are, by default, serialized to string using JSONB before being passed down to the store.

* **What is the new behavior (if this is a feature change)?**
* Instead of expecting a Store<string | undefined> , expect Store<Value>

* **Does this PR introduce a breaking change?** (What changes might users need to make in their application due to this PR?)
* No

* **Other information**:


packages/etcd/package.json
```diff
@@ -45,13 +59,15 @@
 	"dependencies": {
 		"cockatiel": "^2.0.2",
 		"etcd3": "^1.1.0",
-		"pify": "^5.0.0"
+		"pify": "^5.0.0",
+		"@types/keyv": "^3.1.4"
```
jaredwray: @alphmth - this should be moved to devDependencies

jaredwray (changes requested):
can you update the dependencies across all of these?

jaredwray (approved):

PR #501, opened 2022-09-27
fixed namespace clear issue

### LINKED ISSUE #455: Redis Fix - namespace key can grow to GB size
**Describe the bug**
Since I migrated to `keyv` it looks like the amount of freeable memory is declining in constant rate: https://imgur.com/a/77kOTD7 (20 Jul is when I released changes for `keyv` migration).

I expect it to be related to the `namespace:...` key, that in my case can grow to more than 15GB in actively used cache.

**To Reproduce**
I would expect that adding a bunch of big keys for short living values will cause the `namespace:...` size to constantly grow.

**Expected behavior**
I would expect to have some auto-purging mechanism?

**Tests (Bonus!!!)**
N/A

**Additional context**
N/A

alphmth:

fixes #455

### DIFF
--- a/packages/redis/src/index.js
+++ b/packages/redis/src/index.js
@@ -68,7 +68,7 @@ class KeyvRedis extends EventEmitter {
 
 	clear() {
 		return this.redis.smembers(this._getNamespace())
-			.then(keys => this.redis.del([...keys, ...this._getNamespace()]))
+			.then(keys => this.redis.del([...keys, this._getNamespace()]))
 			.then(() => undefined);
 	}
 
--- a/packages/redis/test/test.js
+++ b/packages/redis/test/test.js
@@ -106,10 +106,10 @@ test('namespace is not cleaned up by TTL', async t => {
 	const expiredValue = await redis.get(`v3:${key}`);
 	t.is(expiredValue, null);
 
-	// But the namespace doesn't expire
+	//namespace should also expire
 	t.true(await redis.exists('namespace:v3') === 0);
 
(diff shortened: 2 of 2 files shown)


alphmth:
@jaredwray namespace issue for redis is fixed

jaredwray:
referencing this issue: https://github.com/jaredwray/keyv/issues/455

@kdybicz - do you want to validate this fix works and has no issues?

kdybicz:
@jaredwray please correct me if I'm wrong, but I believe this fix will remove the namespace only when `await keyv.clear();` is called. If so, it will still not prevent it to grow in size over time.

In other words:
1. Test name doesn't correspond to what's really tested.
2. Fix does resolve different issue - I'm not questioning importance of that fix here.
3. Proposed solution doesn't introduce any "automated" way for removing expired members from the namespace key, which is the root cause of the growing memory usage.

Hope above points make sense.

jaredwray:
@kdybicz - the question that we will have is if it is even possible to remove expired members from the namespace key.

kdybicz:
@jaredwray that's my main concern and I suspect it might not be possible. In such case it might make sense to simplify the test, just to make sure that `await keyv.clear()` does its job and maybe also add somewhere in the documentation a note that there might be some issues in high availability systems and it such cases it might require to schedule a periodical `clear` to deal with it.

A side question, as I didn't looked that deep into your code: is the namespace always needed?

jaredwray:
going to add this in and then we might want to discuss a better way to handle this in the future where we just dont use namespace.

PR #502, opened 2022-09-27
Compression Updates

jaredwray:

The goal with this is to update the main Keyv to handle compression in a more scalable manner and not do a ton of overwrites. 
- standardize the compression library with `compress(value, options)` and `decompress(value, options)`
- document a better structure for testing with the main Keyv library
- Remove anything unnecessary that was implemented
- Remove `compress` as it is no longer needed.

### DIFF
--- a/packages/compress-brotli/src/index.js
+++ b/packages/compress-brotli/src/index.js
@@ -19,12 +19,12 @@ class KeyvBrotli {
 		};
 	}
 
-	compress(value, options) {
-		return this.opts.compress(value, options);
+	compress(value) {
+		return this.opts.compress(value, this.opts);
 	}
 
-	decompress(value, options) {
-		return this.opts.decompress(value, options);
+	decompress(value) {
+		return this.opts.decompress(value, this.opts);
 	}
 }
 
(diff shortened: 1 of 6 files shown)


packages/compress-brotli/src/index.js
```diff
@@ -19,12 +19,12 @@ class KeyvBrotli {
 		};
 	}
 
-	compress(value, options) {
-		return this.opts.compress(value, options);
+	compress(value) {
```
stropicall, line 22: @jaredwray we need this `options` argument in `compress` or `decompress` methods as `brotli` and `gzip` also have those options filed.

https://github.com/Kikobeats/compress-brotli/blob/master/src/index.js#L29
https://github.com/nodeca/pako/blob/master/lib/inflate.js#L101

jaredwray: @alphmth - the goal is that you would pass the options via the constructor for this compression option. 

in the constructor we go with 

```js
	constructor(options) {
		this.opts = options;
		this.compressBrotli = compressBrotli(options);
	}
```
then we are able to use that for each of the calls. 
```js
	async decompress(value) {
		const {value, expires} = this.compressBrotli.deserialize(data);
		return {value: await this.compressBrotli.decompress(value), expires};
	}
```

PR #511, opened 2022-10-03
Compression updates v2

alphmth:

compression updates


packages/compress-brotli/src/index.js
```diff
@@ -3,28 +3,25 @@ const compressBrotli = require('compress-brotli');
...
-		};
-
-		const {compress, decompress, serialize, deserialize} = compressBrotli(this.opts);
+		this.opts = options;
+		this.brotli = compressBrotli(this.opts);
+	}
 
-		this.opts.compress = compress;
-		this.opts.serialize = async ({value, expires}) => serialize({value: await compress(value), expires});
+	compress(value) {
```
jaredwray: can we make this async?

packages/compress-brotli/src/index.js
```diff
@@ -3,28 +3,25 @@ const compressBrotli = require('compress-brotli');
...
+		return this.brotli.compress(value, this.opts);
+	}
 
-		this.opts.decompress = decompress;
-		this.opts.deserialize = async data => {
-			const {value, expires} = deserialize(data);
-			return {value: await decompress(value), expires};
-		};
+	decompress(data) {
```
jaredwray: can we make this async?

packages/compress-gzip/src/index.d.ts
```diff
@@ -1,19 +1,11 @@
...
-declare namespace KeyvGzip {
-	interface Options {
-		compress: (...args: any[]) => void;
-		decompress: (...args: any[]) => void;
-	}
+	constructor(options?);
+	compress(value: any);
+	decompress(value: any);
+	serialize(value: any);
+	deserialize(value: any);
```
jaredwray: please make these async?

packages/compress-gzip/src/index.js
```diff
@@ -7,23 +7,23 @@ class KeyvGzip {
...
 
-		this.opts.compress = pako.deflate;
-		this.opts.decompress = pako.inflate;
-		this.opts.serialize = async ({value, expires}) => JSONB.stringify({value: await this.opts.compress(value, this.opts), expires});
-		this.opts.deserialize = async data => {
-			const {value, expires} = JSONB.parse(data);
-			const value_ = await this.opts.decompress(value, this.opts);
-			return {value: value_, expires};
-		};
+	compress(value) {
```
jaredwray: async please

packages/compress-gzip/src/index.js
```diff
@@ -7,23 +7,23 @@ class KeyvGzip {
...
+		return pako.deflate(value, this.opts);
+	}
+
+	decompress(value) {
```
jaredwray: async please

packages/keyv/src/index.d.ts
```diff
@@ -85,7 +85,12 @@ declare namespace Keyv {
 		/** Specify an adapter to use. e.g `'redis'` or `'mongodb'`. */
 		adapter?: 'redis' | 'mongodb' | 'mongo' | 'sqlite' | 'postgresql' | 'postgres' | 'mysql' | undefined;
 		/** Enable compression option **/
-		compress?: Record<string, unknown> | undefined;
+		compression?: CompressionAdapter | undefined;
+	}
+
+	interface CompressionAdapter {
+		compress: (value: any) => any;
+		decompress: (value: any) => any;
```
jaredwray: can we add in `options` parameter?

jaredwray (changes requested):

packages/keyv/src/index.d.ts
stropicall: @jaredwray Earlier we are supporting `options` parameter but you said pass the `options` parameter in constructor

jaredwray: I would like it to be passed via a property or as a parameter in the constructor. Either way it should work and we should test for that.

stropicall: ok @jaredwray I will do that

alphmth:
> @alphmth - how would we separate this out so that serialize, deserialized could exist without being part of the compression package or maybe an option that if those functions exist we use it otherwise we use what is set?

In the `set` or `get` method we are using `serialize` and `deserialize` methods. To make compression work we need to add compress/decompress method in the serialize/deserialize method. If we make them optional how we will use compression in main `keyv` project?

jaredwray:
@alphmth - this is starting to come together. One other thought is to create another export of the test-suite library that does validation of the compression packages so third parties can do that. Can you add that in?

alphmth:
@jaredwray do we need to add `serialize/deserialize` as optional? In that case, the user will have to use `compress/decompress` methods like this

```
const gzip = new KeyvGzip(); 
const compressedValue = gzip.compressed(value);
const keyv = new Keyv({store: new Map()});
await keyv.set('test-key', compressedValue, 1000);

//get compressed value
const decompressedValue = await gzip.decompressed(await keyv.get('test-key));

```

jaredwray:
> @jaredwray do we need to add `serialize/deserialize` as optional? In that case, the user will have to use `compress/decompress` methods like this
> 
> ```
> const gzip = new KeyvGzip(); 
> const compressedValue = gzip.compressed(value);
> const keyv = new Keyv({store: new Map()});
> await keyv.set('test-key', compressedValue, 1000);
> 
> //get compressed value
> const decompressedValue = await gzip.decompressed(await keyv.get('test-key));
> ```

Lets go with the current architecture for right now and then will take feedback.

Issue #527, opened 2022-10-18
Allow passing SSL configuration into options for @key/postgres


wgwz:
I recently found that I was unable to use the keyv postgres backend because the production database I was connecting to forces SSL connections. This is a request to add the ability to pass SSL configuration options into the Keyv object. Or more generally to allow passing external pool for Keyv to make use of.

These are a sample of the errors I was getting:

```
Error from keyv.Keyv: error: no pg_hba.conf entry for host "3.89.49.50", user "postgres", database "postgres", SSL off
```

And the more cryptic message (this occurs after the above error):

```
TypeError: query is not a function
  File "/app/node_modules/@keyv/postgres/src/index.js", line 29, col 19, in <anonymous>
    .then(query => query(sqlString, values));
  File "/app/node_modules/@keyv/postgres/src/index.js", in runMicrotasks
  File "node:internal/process/task_queues", line 96, col 5, in processTicksAndRejections
```

Here's a sample from my project of the SSL options:

```typescript
import * as fs from 'fs';
import { Pool, PoolConfig } from 'pg';

function setupPgPool(): Pool {
  const poolConfig: PoolConfig = {
    connectionString:
      process.env.DATABASE_URL ||
      'postgres://postgres:postgres@localhost:5432/regen_registry',
  };

  if (process.env.NODE_ENV === 'production') {
    poolConfig.ssl = {
      ca: fs.readFileSync(`${__dirname}/../config/rds-combined-ca-bundle.pem`),
    };
  }

  const pool = new Pool(poolConfig);
  return pool;
}

const pgPool = setupPgPool();

export { pgPool };
```

Here's an [upstream issue](https://github.com/regen-network/regen-registry/issues/1311) with some additional info in case it is helpful. I would be glad to work on this once an approach is decided.

jaredwray:
@wgwz - Thanks for sending this over. It looks like we are not testing for SSL at this time and would be interested in adding it. Also, it looks like the link https://github.com/regen-network/regen-registry/issues/1311 is private or no longer working. 

Are you interested in doing a pull request?If not, more than happy to leave this for a vote to see if this will get any traction.

wgwz:
hey @jaredwray thanks for the quick reply, and yes, i'm interested in working on a PR!

sorry about that link, i forgot it was private, but i did add most of the important info in the description. our tracker may end up public eventually, so we can just leave the link for now.

jaredwray:
[...]

PR #529, opened 2022-10-19
Allow passing ssl config options for @keyv/postgres

wgwz:

Closes: https://github.com/jaredwray/keyv/issues/527

Current status:
- docker-compose service added for running postgres in ssl-only mode
- new test file added for SSL tests (expectation is that they all will fail right now)

Next steps:
- set up passing ssl options in KeyvPostgres
- feedback please!


jaredwray:
@wgwz - looking good so far. Thanks for working on this!

wgwz:
These postgres tests are passing locally, with the changes made in [211edbe](https://github.com/jaredwray/keyv/pull/529/commits/211edbe228b126a3c7781412559d9fb83b8aa29e). Though when I ran the full test suite I had some tests from etcd that were failing, it had some message related to grpc.

One thing that I tried to add but could not get working is ssl certificate validation (to copy what AWS RDS does for example, see "verify-ca"). https://www.postgresql.org/docs/current/libpq-ssl.html. In other words, I was aiming to have a test case where you have to pass some "ca" option in the SSL options.

But then I thought, it'd rather be me testing that node-postgres is doing it's job, since this implementation really is just passing the SSL options to the Pool constructor. LMK what you think. 

I did test locally with a connection to an RDS database that requires verify-ca, and it's working as expected.

jaredwray:
@wgwz - Looks like the tests are failing to connect to the ssl instance after spinning it up with docker compose. 

```bash
    pg_ctl -D /var/lib/postgresql/data -l logfile start

waiting for server to start....2022-10-24 11:41:57.663 UTC [47] FATAL:  private key file "/var/lib/postgresql/server.key" has group or world access
2022-10-24 11:41:57.663 UTC [47] DETAIL:  File must have permissions u=rw (0600) or less if owned by the database user, or permissions u=rw,g=r (0640) or less if owned by root.
initdb: warning: enabling "trust" authentication for local connections
You can change this by editing pg_hba.conf or using the option -A, or
--auth-local and --auth-host, the next time you run initdb.
pg_ctl: could not start server
Examine the log output.
2022-10-24 11:44:20.521 UTC [1] FATAL:  private key file "/var/lib/postgresql/server.key" has group or world access
2022-10-24 11:44:20.521 UTC [1] DETAIL:  File must have permissions u=rw (0600) or less if owned by the database user, or permissions u=rw,g=r (0640) or less if owned by root.
2022-10-24 11:44:20.521 UTC [1] LOG:  database system is shut down
2022-10-24 11:41:57.663 UTC [47] LOG:  database system is shut down
 stopped waiting
```
I haven't done much work at setting up postgres with SSL. Do you have a reference doc or guide that I can look at to see if I can help get this running?

wgwz:
@jaredwray i found the postgresql guide to be pretty complete:

https://www.postgresql.org/docs/current/ssl-tcp.html

for this issue though, we just have to adjust the permissions of that key, which explains why these are passing in my local vs not in CI. maybe we can do that through the docker-compose file mounts? we could also write a small custom dockerfile.

packages/postgres/test/test-ssl.js
```diff
@@ -0,0 +1,63 @@
...
+		await keyv.get('foo');
+		t.fail();
+	} catch {
+		t.pass();
+	} finally {
+		endPool();
+	}
+});
+
+const options = {ssl: {rejectUnauthorized: false}};
```
jaredwray, line 21: @wgwz  - is this the only option for the pool on ssl? Are there other options that need to happen for it to work? The reason I am asking is I want to build documentation on the README for others on how to setup Keyv using postgres and SSL option.

jaredwray:
@wgwz - this is looking great. Can you answer the questions above and we can get this checked in?

wgwz:
@jaredwray sorry bout that. somehow the replies i added are stuck in a "pending" state, idk how to fix it. here's my reply:

>is this the only option for the pool on ssl? Are there other options that need to happen for it to work? The reason I am asking is I want to build documentation on the README for others on how to setup Keyv using postgres and SSL option.

there are other options in [node-postgres](https://node-postgres.com/features/ssl). which in turn leads to any [options for a node tls socket](https://nodejs.org/api/tls.html#tls_class_tls_tlssocket). which in turn leads to any [options for createSecureContext](https://nodejs.org/api/tls.html#tlscreatesecurecontextoptions) (it's a bit of a rabbit hole). but the most commonly used would be these i guess (as in the node-postgres docs):

```
  ssl: {
    rejectUnauthorized: false,
    ca: fs.readFileSync('/path/to/server-certificates/root.crt').toString(),
    key: fs.readFileSync('/path/to/client-key/postgresql.key').toString(),
    cert: fs.readFileSync('/path/to/client-certificates/postgresql.crt').toString(),
  },
```

`rejectUnauthorized` is a development convenience, which should probably not be recommended in production. for the codebase i'm using this in, we only need to specify the `ca` option. we're connecting to an AWS RDS postgres database with `verify-ca` enabled. here's more docs on [AWS postgres RDS ssl](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/PostgreSQL.Concepts.General.SSL.html) in case it's helpful.

in the case for the tests we have here, we only need the `rejectUnauthorized` option. because there's a mismatch in the common name of the certificate and localhost or something like that. there's probably a way that it could be fixed, and then we could just change that to `true`. but it's not too big of an issue i think for the sake of these tests.

i'm thinking that because the underlying node-postgres seems to do the same type of deferral to the node tls primitives, it makes sense to do the same here, and not worry about testing all the dimensions too much.

jaredwray:
@wgwz - thank you so much! We will be updating documentation and looking for a release the week of November 6th. Will merge this in between then and get it ready!

PR #722, opened 2023-03-07
tiered: add typescript support

christianllv:


packages/tiered/src/index.ts
```diff
@@ -0,0 +1,119 @@
+import EventEmitter from 'node:events';
+import Keyv from 'keyv';
+import type {Options, Options_} from './types';
```
jaredwray, line 3: What is `Options_` vs `Options` ?

packages/tiered/src/index.ts
```diff
@@ -0,0 +1,119 @@
...
+		this.opts = {
+			validator: () => true,
+			dialect: 'tiered',
+			...options,
+		};
+		this.remote = remote;
+		this.local = local;
+	}
+
+	async get(key: string): Promise<any> {
```
jaredwray, line 24: Do we want to return `any` or could we define this more?

jaredwray (changes requested):

packages/tiered/src/index.ts
chrisllontop: Options is the parameters that we receive in the constructor. Options_ are the parameters in this.opts

packages/tiered/src/index.ts
chrisllontop: We don't know the data type cached, for this reason, we can return any.

jaredwray (approved):

PR #728, opened 2023-03-08
typings: Correctly type `Keyv.get([])` signature to include `undefined` return type

### LINKED ISSUE #720: Plural `Keyv.get([])` types are incorrect, or `Store.getMany()` returns incorrect type
Additional context: https://github.com/apollographql/apollo-utils/pull/260

**Describe the bug**
The "plural" `Keyv.get([])` overload is expected to return `Promise<Array<V | undefined>>` based on the typings. However, you can see in my PR that when a `Store.getMany()` returns `undefined`, then `undefined` is returned by `Keyv.get([])`.

Should `Store.getMany()` be required to always return an array (I think no, `undefined` seems semantically useful in the error case, maybe)? Or should the `Keyv.get([])` types be updated to include the possibility of `undefined` so we know to handle that case?

**How To Reproduce (best to provide workable code or tests!)**
Reproduction available in the linked PR.
The initial commit demonstrating bad behavior might be more interesting:
https://github.com/apollographql/apollo-utils/pull/260/commits/60621069d3d1dee29905ccc29af199194b73fc77

trevor-scheer:
Fixes #720

This change introduces a "bug fix" to the typings of the `Keyv.get([])` signature. Because `Store.getMany()` can return `undefined`, `Keyv.get([])` should include that in its possible return types.

This does mean that consumers of these types will incur TS errors on their plural `get()` usages when they upgrade / install this patch, but those errors have always existed in runtime so I would consider this a fix which aligns the types with reality.

Demo of current bad behavior: https://codesandbox.io/p/sandbox/typescript-node-forked-dpx2ge

### DIFF
--- a/packages/keyv/src/index.d.ts
+++ b/packages/keyv/src/index.d.ts
@@ -37,7 +37,7 @@ declare class Keyv<Value = any, Options extends Record<string, any> = Record<str
 		key: string[],
 		options?: {raw?: Raw}
 	): Promise<
-	Array<(Raw extends false ? Value : Keyv.DeserializedData<Value>) | undefined>
+	Array<(Raw extends false ? Value : Keyv.DeserializedData<Value>) | undefined> | undefined
 	>;
 
 	/**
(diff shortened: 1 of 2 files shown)


jaredwray:
@trevor-scheer - do you know what is missing on the tests to get the code coverage back to 100%?

trevor-scheer:
@jaredwray rebased and fixed codecov back to 100%, though I'm not sure why my original change caused a drop.

PR #827, opened 2023-05-25
fix: maximum call stack size exceeded

AuspeXeu:
Unfortunately, if the redis database has many entries, the current approach using recursion can run into a maximum call stack size exceeded exception.

image

### DIFF
--- a/packages/redis/src/index.ts
+++ b/packages/redis/src/index.ts
@@ -81,27 +81,21 @@ class KeyvRedis<Value = any> extends EventEmitter {
 	}
 
 	async * iterator(namespace?: string): IteratorOutput {
-		const scan = this.redis.scan.bind(this.redis);
-		const get = this.redis.mget.bind(this.redis);
-		// @ts-expect-error - iterator
-		async function * iterate(curs, pattern) {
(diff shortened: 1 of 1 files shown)


lzfxxx:
@jaredwray I've also encountered the same issue that's been addressed by this Pull Request #827 from @AuspeXeu. The problem with maximum call stack size exceeded exception due to recursion when the redis database has many entries is affecting my project as well. I kindly request that this Pull Request be considered for merging at the earliest possible convenience to resolve this common issue. I appreciate your work on this project and thank you in advance for your attention to this matter.

jaredwray:
@AuspeXeu and @lzfxxx - thanks for the pull request. Can you get this updated as we just converted Redis to typescript. Can you do it based on that and make sure the tests run successfully so I can validate?

AuspeXeu:
@jaredwray the tests fail due to a linting rule. Which does not seem to make sense for this particular case. Maybe we can deactivate it?

jaredwray:
@AuspeXeu and @lzfxxx this should be released next weekend with other updates. Thanks so much for the pull request!!!

PR #855, opened 2023-06-22
add useRedisSets option  

christianllv:

### DIFF
--- a/packages/redis/README.md
+++ b/packages/redis/README.md
@@ -68,6 +68,30 @@ const keyvRedis = new KeyvRedis(redis);
 const keyv = new Keyv({ store: keyvRedis });
 ```
 
+## Options
+
+### useRedisSets
+
+The `useRedisSets` option lets you decide whether to use Redis sets for key management. By default, this option is set to `true`.
+
+When `useRedisSets` is enabled (`true`):
+
+- A namespace for the Redis sets is created, and all created keys are added to this. This allows for group management of keys.
+- When a key is deleted, it's removed not only from the main storage but also from the Redis set.
+- When clearing all keys (using the `clear` function), all keys in the Redis set are looked up for deletion. The set itself is also deleted.
+
+If you decide to set `useRedisSets` as `false`, keys will be handled individually and Redis sets won't be utilized.
+
+#### Example
+
+Here's how you can use the `useRedisSets` option:
+
+```js
+const Keyv = require('keyv');
+
+const keyv = new Keyv('redis://user:pass@localhost:6379', { useRedisSets: false });
+```
+
 ## License
 
 MIT © Jared Wray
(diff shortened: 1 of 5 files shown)


jaredwray:
Also, will we be able to support setting this option on the main Keyv such as this:

```javascript
const keyv = new Keyv('redis://user:pass@localhost:6379', {useRedisSets: false});
```

I would also assume that by default we are using the Redis Sets to make sure this is backwards compatible?

christianllv:
Hello @jaredwray we are using only sets (without sadd) only when useredisSets is true. But, by default is false. Should I set useRedisSets in true by default?

christianllv:
Also, yes we can use that option as option in keyv

jaredwray:
> Hello @jaredwray we are using only sets (without sadd) only when useredisSets is true. But, by default is false. Should I set useRedisSets in true by default?

I am thinking for backwards compatibility we keep have `useRedisSets` to `true` as that is what it is today.

jaredwray:
> Also, yes we can use that option in keyv

Can you update the Redis README with this new option and examples?

christianllv:
@jaredwray docs added. Please take a look. Thanks!

packages/redis/README.md
```diff
@@ -68,6 +68,30 @@ const keyvRedis = new KeyvRedis(redis);
...
+
+The `useRedisSets` option lets you decide whether to use Redis sets for key management. By default, this option is set to `true`.
+
+When `useRedisSets` is enabled (`true`):
+
+- A namespace for the Redis sets is created, and all created keys are added to this. This allows for group management of keys.
+- When a key is deleted, it's removed not only from the main storage but also from the Redis set.
+- When clearing all keys (using the `clear` function), all keys in the Redis set are looked up for deletion. The set itself is also deleted.
+
+If you decide to set `useRedisSets` as `false`, keys will be handled individually and Redis sets won't be utilized.
```
jaredwray, line 84: Can we put a note that we have seen memory leaks when in high performance scenarios so it is best to turn this to false if in high performance.

packages/redis/README.md
```diff
@@ -68,6 +68,30 @@ const keyvRedis = new KeyvRedis(redis);
 const keyv = new Keyv({ store: keyvRedis });
 ```
```
jaredwray, line 69: It would be good to put a note on `clear()` that if you turn `useRedisSets` to false it could cause performance issues in production as it will need to iterate over all keys to delete them.

packages/redis/README.md
chrisllontop: Please check the docs updated.

packages/redis/README.md
christianllv: Please check the docs updated.

PR #865, opened 2023-07-05
fix: add `disconnect` to memcache driver

Julien-R44:

Without this change, this code was never exiting : 

```ts
import Keyv from 'keyv';
import KeyvMemcache from './src/index';

let uri = 'localhost:11211';
const keyvMemcache = new KeyvMemcache(uri);
const keyv = new Keyv({store: keyvMemcache});

keyv.get('foo').then(() => {
  keyv.disconnect()
})
```

So just added a `disconnect` method to the memcache driver, and now works fine

---

Wanted to add a test for the same : 

```ts
test.serial('close connection successfully', async t => {
	const keyv = new Keyv({store: keyvMemcache});
	t.is(await keyv.get('foo'), undefined);
	await keyv.disconnect();
	try {
		await keyv.get('foo', 34);
		t.fail();
	} catch {
		t.pass();
	}
});
```

but it doesn't work. memjs throws 0 errors when you do a `.get` or `.set` when the connection is closed. So not sure how to handle this

### DIFF
--- a/packages/memcache/src/index.ts
+++ b/packages/memcache/src/index.ts
@@ -163,6 +163,10 @@ class KeyvMemcache<Value = any> extends EventEmitter implements Store<Value> {
 			});
 		});
 	}
+
+	async disconnect(): Promise<void> {
+		this.client.quit();
+	}
 }
 
 export = KeyvMemcache;
--- a/packages/memcache/test/test.ts
+++ b/packages/memcache/test/test.ts
@@ -214,10 +214,13 @@ test.serial('close connection successfully', async t => {
 	/**
 	 * Since the memjs library doesn't throw an error when trying to get or set on a closed connection,
    * we need to set up a "fallback" error that will occur if the operation doesn't complete within a reasonable timeframe.
-	 * 
+	 *
 	 * At least this way we can be sure that calling .disconnect() is really closing the connection
 	 */
-	const delayedFailure = new Promise((_, reject) => setTimeout(() => reject(new Error('Operation timed out')), 3000));
+	const delayedFailure = new Promise((_, reject) => setTimeout(() => {
+		reject(new Error('Operation timed out'));
+	}, 3000));
+
 	try {
 		await Promise.race([keyv.get('foo'), delayedFailure]);
 		t.fail();
--- a/packages/memcache/test/test.ts
+++ b/packages/memcache/test/test.ts
@@ -206,6 +206,26 @@ test('get should emit an error', withCallback(async (t: ExecutionContext<any>, e
 	} catch {}
 }));
 
+test.serial('close connection successfully', async t => {
+	const keyv = new Keyv({store: keyvMemcache});
+	t.is(await keyv.get('foo'), undefined);
+	await keyv.disconnect();
+
+	/**
+	 * Since the memjs library doesn't throw an error when trying to get or set on a closed connection,
+   * we need to set up a "fallback" error that will occur if the operation doesn't complete within a reasonable timeframe.
+	 * 
+	 * At least this way we can be sure that calling .disconnect()
+	 */
+	const delayedFailure = new Promise((_, reject) => setTimeout(() => reject(new Error('Operation timed out')), 3000));
+	try {
+		await Promise.race([keyv.get('foo'), delayedFailure]);
+		t.fail();
+	} catch {
+		t.pass();
+	}
+});
+
 const store = () => keyvMemcache;
 
 keyvApiTests(test, Keyv, store);
--- a/packages/memcache/test/test.ts
+++ b/packages/memcache/test/test.ts
@@ -215,7 +215,7 @@ test.serial('close connection successfully', async t => {
 	 * Since the memjs library doesn't throw an error when trying to get or set on a closed connection,
    * we need to set up a "fallback" error that will occur if the operation doesn't complete within a reasonable timeframe.
 	 * 
-	 * At least this way we can be sure that calling .disconnect()
+	 * At least this way we can be sure that calling .disconnect() is really closing the connection
 	 */
 	const delayedFailure = new Promise((_, reject) => setTimeout(() => reject(new Error('Operation timed out')), 3000));
 	try {


jaredwray:
Hi @Julien-R44 👋 - Thanks so much for sending the pull request. If you could add in the unit tests for this change then we could get this merged.

Julien-R44:
Added a test for this one. 
Also opened an issue here to know if this behavior is by design or just a bug : https://github.com/memcachier/memjs/issues/172

If this is by design, we should absolutely fix it on the keyv side. Look at this code : 

```ts
const memcached = new KeyvMemcache('localhost:11211')
const memcache = new Keyv({ store: memcached })

try {
  console.log('go')
  await memcache.get('foo')
  await memcache.disconnect()
  await memcached.get('foo')
  console.log('done 1')
} catch (err) {
  console.log('done 2')
}

console.log('done 3')
```

Guess the output 😄
Output will be : 

```
go
disconnect
```

Because this promise : https://github.com/JuliensForks/keyv/blob/main/packages/memcache/src/index.ts#L48 is never resolving.
So it's quite a big problem.

( This will only be the case after this PR has been applied. Because right now, the `disconnect()` on the memcache driver does nothing. )

packages/memcache/src/index.ts
```diff
@@ -163,6 +163,10 @@ class KeyvMemcache<Value = any> extends EventEmitter implements Store<Value> {
 			});
 		});
 	}
+
+	async disconnect(): Promise<void> {
+		this.client.quit();
```
jaredwray, line 168: to force the quit you might want to do `close()` as that might resolve the issue you are seeing

jaredwray:
> Added a test for this one. Also opened an issue on memjs to know if this behavior is by design or just a bug : [memcachier/memjs#172](https://github.com/memcachier/memjs/issues/172)
> 
> If this is by design, we should absolutely fix it on the keyv side. Look at this code :
> 
> ```ts
> const memcached = new KeyvMemcache('localhost:11211')
> const memcache = new Keyv({ store: memcached })
> 
> try {
>   console.log('go')
>   await memcache.get('foo')
>   await memcache.disconnect()
>   await memcached.get('foo')
>   console.log('done 1')
> } catch (err) {
>   console.log('done 2')
> }
> 
> console.log('done 3')
> ```
> 
> Guess the output 😄 Output will be :
> 
> ```
> go
> disconnect
> ```
> 
> output will not include `done 1`, `done 2`, or `done 3`
> 
> Because this promise : https://github.com/JuliensForks/keyv/blob/main/packages/memcache/src/index.ts#L48 is never resolving. So it's quite a big problem.
> 
> ( This will only be the case after this PR has been applied. Because right now, the `disconnect()` on the memcache driver does nothing. )

that is interesting. Did you want to provide a fix for this in the library as i would like to not add it if it does nothing at this time.  you could also look at close which might do the trick.

packages/memcache/src/index.ts
Julien-R44: Just tried, and it changes nothing. By reading comments in memjs, i believe it's better to use quit as it is a clean exit 
image

Julien-R44:
Not sure I understand what you're proposing
I'll recap to make sure we understand each other :

- This PR adds a `disconnect()` method, which works fine. It really disconnects us from memcache server.
- On the other hand, introducing this method causes the bug I mentioned just above, which could probably annoy some people using keyv.

For this, we have two solutions: 
- Either we wait for a response from the Memjs maintainers, to find out whether this is a bug or normal behavior. Then we act accordingly. I don't think we'll get an answer for a while, as memjs doesn't seem to be extremely maintained. 
 
- Or we could add a fix on the keyv side. The fix on the keyv side would be more or less the same as what I added in the tests. So basically a timeout system. if there's no response after X seconds, then we resolve the various promises with an error.
This way, the program can continue to run.

let me know what you think is best and i'll do a PR

packages/memcache/src/index.ts
jaredwray: ok. thanks for trying. Should we wait to see what memjs says?

jaredwray:
> Not sure I understand what you're proposing I tried `quit()`, but it doesnt changes anything I'll recap to make sure we understand each other :
> 
> * This PR adds a `disconnect()` method, which works fine. It really disconnects us from memcache server.
> * On the other hand, introducing this method causes the bug I mentioned just above, which could probably annoy some people using keyv.
> 
> For this, we have two solutions:
> 
> * Either we wait for a response from the Memjs maintainers, to find out whether this is a bug or normal behavior. Then we act accordingly. I don't think we'll get an answer for a while, as memjs doesn't seem to be extremely maintained.
> * Or we could add a fix on the keyv side. The fix on the keyv side would be more or less the same as what I added in the tests. So basically a timeout system. if there's no response after X seconds, then we resolve the various promises with an error.
>   This way, the program can continue to run.
> 
> let me know what you think is best and i'll do a PR

If you want to do the fix we can support that. If not, lets wait a week to see what they say.

jaredwray:
@Julien-R44  looks like there was a comment. https://github.com/memcachier/memjs/issues/172#issuecomment-1630622740

Thoughts on that?

Julien-R44:
Yeah, I'll try to send a PR to memjs and then update our PR accordingly once the problem has been fixed
Thanks

jaredwray:
> Yeah, I'll try to send a PR to memjs and then update our PR accordingly once the problem has been fixed Thanks

Any word on the PR for memjs? Anything I can do to help?

Julien-R44:
hey! i am sorry to leaving this hanging. i started looking but it's a bit complicated. it's very oldschool js code and a bit low level. 
i'll have to get back into it when i have a bit more time.
but it's still relevant. we're probably planning to use keyv for the caching module in the next version of [AdonisJS](https://github.com/adonisjs/core)

Julien-R44:
sorry, I decided to not add memcached support to my application after all. So unfortunately I won't be able to complete this PR. I'll let you decide whether to close it or not

jaredwray:
> sorry, I decided to not add memcached support to my application after all. So unfortunately I won't be able to complete this PR. I'll let you decide whether to close it or not

@Julien-R44 - thanks for the update and we will look to see if we want to continue on this PR.

jaredwray:
closing and will look at this later

Issue #924, opened 2023-09-29
Check the namespace provided to Redis doesn not include ":"


markNZed:
**Describe the bug**
I had prefixed a namespace like "something:somethingelse" and the Redis iterator was returning the key and including the namespace. I am not sure if this is a bug (should keyv support `:` in the namespace) or a limitation on Redis and an assertion/error might help.

**How To Reproduce (best to provide workable code or tests!)**

jaredwray:
@markNZed - can you provide a unit test example on how to reproduce?

markNZed:
I can have a go at it. Do you want a PR with a test added to https://github.com/jaredwray/keyv/blob/main/packages/test-suite/src/iterator.ts ?

jaredwray:
That would be great. Let me know if you have any questions.

markNZed:
I like that you have the docker environment to run the tests. I had issues with node/yarn on the server (normally I'm running from inside a docker container on that machine). I wonder whether the yarn/node requirement could be removed so we only need docker and then run the tests inside a docker container. It is not so obvous for an inexperienced dev like me to understand how to modify the test-suite and rerun it in a particular context e.g. redis. Basically I needed to modify the source in packages/test-suit then run yarn build then run yarn test in packages/redis. it is hard to know how much explanation should be provided, I am not familiar with TS or Yarn but I guess most people who would do a PR would b familiar with those tools. https://github.com/jaredwray/keyv/pull/926

jaredwray:
```

	test.serial(
		'iterator() handles : in keys X',
		async t => {
			const keyvStore = store();

			const keyv1 = new Keyv({store: keyvStore, namespace: 'keyv1:testv1'});
			const map1 = new Map(
				Array.from({length: 5})
					.fill(0)
					.map((_x, i) => [String(i) + ":" +	String(i), String(i + 10)]),
			);
			const toResolve = [];
			for (const [key, value] of map1) {
				toResolve.push(keyv1.set(key, value));
			}

			await Promise.all(toResolve);

			t.plan(map1.size);
			for await (const [key, value] of keyv1.iterator()) {
			 	console.log("key", key, "value", value);
				const doesKeyExist = map1.has(key);
				const isValueSame = map1.get(key) === value;
				t.true(doesKeyExist && isValueSame);
			}
		},
	);

};
```

jaredwray:
[...]

PR #926, opened 2023-10-05
namespace with : causing problem with iterator

markNZed:
- [X] Followed the [Contributing](https://github.com/jaredwray/keyv/blob/main/CONTRIBUTING.md) guidelines.
- [X] Tests for the changes have been added (for bug fixes/features) with 100% code coverage.
- [X] Docs have been added / updated (for bug fixes / features)

This adds a failing test that shows how ':' in the namespace causes the Redis iterator to fail.

There is some dumping to the console to see the issue (this is from running the test in the @keyv/redis package):
```
key testv1:3:3 value 13
  ✘ [fail]: test › iterator() handles : in keys X
key testv1:2:2 value 12
key testv1:4:4 value 14
key testv1:1:1 value 11
key testv1:0:0 value 10
```

This is in response to the issue https://github.com/jaredwray/keyv/issues/924

### DIFF
--- a/packages/test-suite/src/iterator.ts
+++ b/packages/test-suite/src/iterator.ts
@@ -102,6 +102,35 @@ const keyvIteratorTests = (test: TestFn, Keyv: typeof KeyvModule, store: KeyvSto
 			t.is(entry.value, undefined);
 		},
 	);
+
+	test.serial(
+		'iterator() handles : in keys X',
+		async t => {
+			const keyvStore = store();
+
+			const keyv1 = new Keyv({store: keyvStore, namespace: 'keyv1:testv1'});
+			const map1 = new Map(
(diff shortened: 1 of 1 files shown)


jaredwray:
@markNZed - I am wondering if we should hold off on this fix as there are plans to make namespace a function in the next 3-4 months. Thoughts?

markNZed:
All I provided in the PR is the test case, not a fix. Maybe that is not what is intended for PR ? I agree that it makes sense to wait for the planned changes rather than fixing code that will be replaced. Still I would add this test. Maybe just leave the PR open and copy the test when the new code is available?

jaredwray:
@markNZed - will close this out as the pull request is used to fix any issues. Will move your code example to the issues as we can look at it there.

PR #1173, opened 2024-10-24
fix: preDelete hook not called when a single key is deleted

ugo-studio:
I was using the package and noticed that preDelete is not called, and postDelete doesn't return the keys that where deleted.
I just added proper calling of the hooks


jaredwray:
@ugo-studio - thanks so much for submitting this PR! We should have this published in the next week or two!

Issue #1222, opened 2024-12-02
Ambiguous behavior of `clear` method with no namespace in Redis adapter


mahdavipanah:
**Description**
When reading through the test cases for the Redis adapter I came across this test case (on [this](https://github.com/jaredwray/keyv/blob/4e59d1c5cd65b505b9a38dc7b20e6c8a4f7b9264/packages/redis/test/test.ts#L265) line) that says:

```typescript
test('should clear with no namespace but not the namespace ones', async () => {
    const keyvRedis = new KeyvRedis();
    const client = await keyvRedis.getClient() as RedisClientType;
    await client.flushDb();
    keyvRedis.namespace = 'ns1';
    await keyvRedis.set('foo91', 'bar');
    keyvRedis.namespace = undefined;
    await keyvRedis.set('foo912', 'bar2');
    await keyvRedis.set('foo913', 'bar3');
    await keyvRedis.clear();
    keyvRedis.namespace = 'ns1';
    const value = await keyvRedis.get('foo91');
    expect(value).toBe('bar');
    await keyvRedis.disconnect();
});
```
which matches the implementation meaning that if the namespace is not set the `clear` method will not remove any keys with the `keyPrefixSeparator` in it. This looks very surprising as it's not mentioned anywhere in the README. Actually, there is this description in the README that to me indicates a totally different behavior, which is that the `clear` method will remove **all keys** if no namespace is set:
 
> **clear** - Clear all keys. If the namespace is set it will only clear keys with that namespace.

**Proposal**
Options:
1. Change the behavior of the `clear` method so it removes all keys (basically a `FLUSHDB`) when no namespace is set
2. Update the README file so it makes this behavior very clear

I am in favor of _option 1_ for these reasons:
1. It is more intuitive to think that when no namespace is set, all the defined keys in the DB are in the space of the Keyv instance
2. It makes the clear method useful in certain production situations, but currently due to its logic it's very risky to use it in production

jaredwray:
@mahdavipanah - the issue that we have is the lack of native namespace support in redis without using something like Redis Sets which causes major performance issues. I see both scenarios that we should support:

* if you are using multiple namespaces but then switch to undefined it should only clear the keys that are not appended with a namespace -- this is currently supported but the documentation needs to be updated. 
* if you want to do a complete clear (`flushDb()`) by setting the namespace to `undefined` then we should add in a option such as `flushDbOnClear` which you can set it and the namespace is `undefined` it will do a full clear. 

[...]

PR #1243, opened 2024-12-10
feat: add noNamespaceAffectsAll option

mahdavipanah:

**What kind of change does this PR introduce?**
This PR is a follow-up on issue #1222 which I reported earlier. It adds an option to the adapter's API called `noNamespacesAffectsAll` and if is set to `true` and `namespace` is undefined:
- `clear` method will use `FLUSHDB` command
- `iterator` method will return all keys regardless of having a namespace prefix

The behavior of these methods will remain the same in cluster mode.
This option has a default value of `false` for backward compatibility so this PR does not introduce a breaching change.

### DIFF
--- a/packages/redis/test/test.ts
+++ b/packages/redis/test/test.ts
@@ -3,7 +3,7 @@ import {
 } from 'vitest';
 import {createClient, type RedisClientType} from 'redis';
 import {delay} from '@keyv/test-suite';
-import KeyvRedis, {createKeyv, createCluster} from '../src/index.js';
+import KeyvRedis, {createKeyv} from '../src/index.js';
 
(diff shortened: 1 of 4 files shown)


packages/redis/test/test.ts
```diff
@@ -3,7 +3,7 @@ import {
 } from 'vitest';
 import {createClient, type RedisClientType} from 'redis';
 import {delay} from '@keyv/test-suite';
-import KeyvRedis, {createKeyv, createCluster} from '../src/index.js';
```
mahdavipanah, line 6: Removed this unused import.

jaredwray:
@mahdavipanah - thanks for doing this. I should have this released in the next couple weeks.

PR #1275, opened 2025-01-14
feat: sqlite - add ttl support

Sadzurami:

Added internal ttl support for sqlite storage adapter.
Also, improved types in some places.

I'm still not sure should we check expired values at 'has' method or not.

### DIFF
--- a/packages/keyv/README.md
+++ b/packages/keyv/README.md
@@ -235,16 +235,18 @@ keyv.deserialize = undefined;
 
 The official storage adapters are covered by [over 150 integration tests](https://github.com/jaredwray/keyv/actions/workflows/tests.yaml) to guarantee consistent behaviour. They are lightweight, efficient wrappers over the DB clients making use of indexes and native TTLs where available.
 
-Database | Adapter | Native TTL
----|---|---
-Redis | [@keyv/redis](https://github.com/jaredwray/keyv/tree/master/packages/redis) | Yes
-Valkey | [@keyv/valkey](https://github.com/jaredwray/keyv/tree/master/packages/valkey) | Yes
-MongoDB | [@keyv/mongo](https://github.com/jaredwray/keyv/tree/master/packages/mongo) | Yes 
-SQLite | [@keyv/sqlite](https://github.com/jaredwray/keyv/tree/master/packages/sqlite) | No 
-PostgreSQL | [@keyv/postgres](https://github.com/jaredwray/keyv/tree/master/packages/postgres) | No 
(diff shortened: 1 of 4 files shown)


Sadzurami:
Also, i guess would be good to add some pragmas before init

like that
```
PRAGMA main.synchronous = NORMAL;
PRAGMA main.journal_mode = WAL2;
PRAGMA main.auto_vacuum = INCREMENTAL;
```

jaredwray:
@Sadzurami - thanks for doing this but I am hesitant to do this as we are introducing timers into the code to do clean up on expiration instead of lazy clearing. This is not a good way to add it in as a piece of functionality and might be better to add this as an option to enable if you want it. Thoughts?

Right now, I would not want to do this without it being something as an extended option and if we did do that I would add in some code to make it more scalable:

* Adding in timer infrastructure to be able to set the timer. You can see what we have done with [CacheableMemory](https://github.com/jaredwray/cacheable/blob/main/packages/cacheable/src/memory.ts) and checkInterval. 
* Most likely extending the table with expires instead of another table would make more sense but we would need backwards compatibility. Even better if we can query on the json as to not change the table schema.

Sadzurami:
> This is not a good way to add it in as a piece of functionality and might be better to add this as an option to enable if you want it.

Make sense, if we want to split functionality by native and internal ttls in future.
For native storages - ttl enabled by default, for internal - optional.

If we want to simplify things, i guess there is no need to make feature optional - it must be enabled, or don't make internals ttl at all.

> Adding in timer infrastructure to be able to set the timer.
Good point, thats reduce number of requests to db. Will do it, if we settle things above.

> Even better if we can query on the json as to not change the table schema.

Agree, but if i understand correctly (maybe not), you suggest smth like that, instead of changing table schema

```text
key - string
value - json ( { "expires": 12345, "value": "actual value" } )
```

If that it, than it might work for keyv-instanses, but will not backwards compatibility with older versions and other apps.
In that case, better to create another table, smth like `${table.name}-ttl` where will be key-ttl pairs for main table.
It will reduce perfomance a bit, since we need to fetch from 2 tables.
But it's okay, if feature with ttl will be optional.

jaredwray:
@Sadzurami -  I think this is a lot of change for the current library. We are planning on doing some work on this library later on in the year and could include things like this, but I don’t know if it’s something that we wanna support now let’s go ahead and close this.

We can look at doing something like cleanup on a check interval later on in the year to see if that’s a feature that we want to add.

PR #1304, opened 2025-02-06
Bugfix: use Object.prototype for correct usage of hasOwnProperty

### LINKED ISSUE #1303: TypeError: data.hasOwnProperty is not a function
**Describe the bug**  
The function `hasOwnProperty` is being called directly on `data` in `packages/serialize/src/index.ts`, which can cause a `TypeError` when `data` is an object created with `Object.create(null)`, or if it overrides `hasOwnProperty` with a non-function value.

This results in the following error:

```
TypeError: data.hasOwnProperty is not a function
```

This issue could lead to crashes when attempting to serialize objects that do not inherit from `Object.prototype`.

---

**How To Reproduce (best to provide workable code or tests!)**  

```js
const data = Object.create(null); // Object without prototype
data.someKey = "value";

function serialize(obj) {
  let s = "";
  const array = Array.isArray(obj);
  s = array ? "[" : "{";
  let first = true;

  for (const k in obj) {
    if (!obj.hasOwnProperty(k)) { // This will throw a TypeError
      continue;
    }
  }

  s += array ? "]" : "}";
  return s;
}

console.log(serialize(data));
```

I'll provide a PR to fix this issue.

nolliebigspin:

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...
fixes #1303

### DIFF
--- a/packages/serialize/src/index.ts
+++ b/packages/serialize/src/index.ts
@@ -28,9 +28,10 @@ export const defaultSerialize = (data: any): string => {
 		// eslint-disable-next-line guard-for-in
 		for (const k in data) {
 			const ignore = typeof data[k] === 'function' || (!array && data[k] === undefined);
-			if (!Object.prototype.hasOwnProperty.call(data, k) || ignore) {
-        continue;
-      }
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-argument
+			if (!Object.hasOwn(data, k) || ignore) {
+				continue;
+			}
 
 			if (!first) {
 				s += ',';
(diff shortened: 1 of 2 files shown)


jaredwray:
@nolliebigspin - thanks for doing this and we will get this fix out in the next week or so.

drodil:
Good finding and fix @nolliebigspin! This same bug was in the `json-buffer` dependency before it was replaced with own serialization (https://github.com/dominictarr/json-buffer/blob/master/index.js#L22)

PR #1399, opened 2025-04-30
fix(redis): apply namespace prefix once and add regression test

ItayLeon-Max:
🛠️ Fix: Namespace Prefix Applied Twice

This PR resolves a bug where the Redis namespace was being prefixed twice when using KeyvRedis with a namespace option.

✅ Changes:
	•	Ensures namespace is only applied once by controlling prefixing in the adapter.
	•	Adds a test to verify users::admin is stored and not users::users::admin.

🧪 Result:

Test namespace-prefix.test.ts now passes as expected:
`expect(keys).toContain('users::admin');
expect(keys).not.toContain('users::users::admin');`

Let me know if you’d like anything adjusted!

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,16 @@
     "xo": "^0.60.0"
   },
   "pnpm": {
-    "onlyBuiltDependencies": ["esbuild", "protobufjs", "sqlite325h", "sqlite3"]
+    "onlyBuiltDependencies": [
+      "esbuild",
+      "protobufjs",
+      "sqlite325h",
+      "sqlite3"
+    ]
+  },
+  "dependencies": {
+    "@redis/client": "^1.6.0",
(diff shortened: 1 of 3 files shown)


ItayLeon-Max:
Hi! Just a quick note:
🛠 This PR was created from a fork, so GitHub workflows are pending your approval (as expected).
More info: [GitHub Docs on approving workflows](https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)

Also, let me know if you’d prefer I remove the console.log debug lines — I only added them to help verify the namespace behavior during testing. I’m happy to push a clean commit without them if needed.

Thanks! 🙏

jaredwray:
@ItayLeon-Max - this seems to not fix anything

PR #1397, opened 2025-04-29
keyv - fix: serialize entries in setMany

nathan-knight:
**Please check if the PR fulfills these requirements**
- [x] Followed the [Contributing](https://github.com/jaredwray/keyv/blob/main/CONTRIBUTING.md) guidelines.
- [ ] Tests for the changes have been added (for bug fixes/features) with 100% code coverage.
- [ ] Docs have been added / updated (for bug fixes / features)

**What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
Bug fix

Currently `setMany` does not serialize the entries currently which causes errors with `@keyv/redis`. I replicated the serialization/ttl fallback logic from `set` for this. I'm currently using this in my application as a patch to get it working.

PR #1504, opened 2025-07-16
Fixed setMany setting entries twice

nathan-knight:

I appreciate you getting my change from #1397 in but you missed a line which made it call `setMany`, then call `set` for each entry anyways and discard the result of calling `setMany`.

One other issue that I uncovered with this change is that `KeyvStoreAdapter.setMany` must return `Promise<void>` so returning the result of `this._store.setMany` won't actually be `boolean[]`.


jaredwray:
@nathan-knight - thanks

Issue #1680, opened 2025-11-06
PGlite storage adapter support


darky:
Hello! Would be great to support [PGlite](https://pglite.dev) as storage adapter
PGlite is useful for unit testing and development 
* On production **postgres** storage adapter can be used
* During development and CI / CD **pglite** storage adapter can be used

Also PGlite can be used on production too as "lightweight" Postgres
BTW, PGlite can be part of current **postgres** storage adapter, almost all current codebase can be reused

jaredwray:
@darky thanks for sending this over. We are doing a big update to postgres and would be interested in adding this. Will add it to something people can vote on to see if it is worth it.

jaredwray:
@darky - not a ton of asks for this and we prefer to use the main server as we have had issues with virtual ones not being fully compatible. Going to close this out but more than happy if you want to make a third party driver for it.

PR #1681, opened 2025-11-08
postgres - chore: connect method as public

darky:
While we are waiting #1680 (if it will of course) here's small cosmetic change for ability to implement own PGlite support like

```ts
class KeyvPglite extends KeyvPostgres {
  override connect() {
    // PGlite connect
  }
  
  override disconnect() {
    // PGlite disconnect
  }
}
```

The rest should works out of the box


jaredwray:
@darky - can we make the functions `public` instead of `protected` for flexibility?

jaredwray:
@darky - thanks!

PR #1695, opened 2025-11-15
postgres - perf: getMany O(n^2) -> O(2n)

darky:


jaredwray:
@darky - thanks so much and I should have this live in the next 8-10 days.

PR #1723, opened 2025-11-18
postgres - feat: setMany support

darky:


jaredwray:
@darky - this should be live in the next week or so. Thanks!

Issue #1805, opened 2026-01-14
KeyvRedis class field initializer creates default Redis connection on module load


lhb-patadata:
**Describe the bug**

The `KeyvRedis` class has a class field initializer `_client = createClient()` that executes when the module is loaded, not when the class is instantiated. This causes an unwanted connection attempt to `localhost:6379` just by importing the `@keyv/redis` package, even if you intend to connect to a different Redis server.

**To Reproduce**

1. Create a file that imports from `@keyv/redis`:
```javascript
import KeyvRedis from '@keyv/redis';
// or
import { createKeyv } from '@keyv/redis';

console.log('Module loaded');
```

2. Run the file without a local Redis server
3. Observe that a connection attempt is made to `localhost:6379` before any code executes

**Expected behavior**

Importing the module should not trigger any network connections. The Redis client should only be created when:
- `new KeyvRedis(url)` is called, or
- `createKeyv(url)` is called

**Root Cause**

In `src/index.ts`, the `KeyvRedis` class has:

```typescript
class KeyvRedis extends Hookified {
  _client = createClient();  // <-- This runs on class definition, not instantiation
  // ...
  
  constructor(connect, options) {
    // The constructor then replaces _client with the correct connection
    if (connect) {
      if (typeof connect === "string") {
        this._client = createClient({ url: connect, socket });
      }
      // ...
    }
  }
}
```

Class field initializers run when the class is defined (during module load), not when `new` is called. This means `createClient()` is called immediately when the module is imported.

**Impact**

This causes issues in several scenarios:

1. **Server startup hangs**: In environments where `localhost:6379` is not available, the default connection attempt can hang indefinitely, blocking application startup.

2. **OpenTelemetry instrumentation**: The `@redis/client` module gets loaded before OpenTelemetry can register its Redis instrumentation, causing instrumentation to fail.

3. **Serverless/Lambda cold starts**: Unnecessary connection attempts slow down cold starts even when Redis isn't needed for that particular request.

4. **Testing**: Unit tests that mock Redis are affected by the real connection attempt during module load.

**Suggested Fix**

Change the class field to not initialize a default client:

```typescript
// Before (problematic)
class KeyvRedis extends Hookified {
  _client = createClient();

// After (fixed)
class KeyvRedis extends Hookified {
  _client: RedisClientType;
  
  constructor(connect, options) {
    super();
    // Initialize _client here based on connect parameter
    // If no connect is provided, then create the default client
    if (!connect) {
[...]

PR #1806, opened 2026-01-14
redis - fix: move client initialization to constructor to prevent connection on module load (#1805)

jaredwray:
The class field initializer `_client = createClient()` was executing at module load time,
causing Redis connection attempts before any KeyvRedis instance was created. This caused
issues with:
- Startup hangs when localhost:6379 is unavailable
- OpenTelemetry instrumentation missing initialization
- Serverless cold start delays
- Testing complications with mock Redis

Moved client initialization from class field to constructor so connections only occur
when explicitly instantiating KeyvRedis or calling createKeyv().

### DIFF
--- a/packages/redis/src/index.ts
+++ b/packages/redis/src/index.ts
@@ -46,8 +46,7 @@ export default class KeyvRedis<T>
 	extends Hookified
 	implements KeyvStoreAdapter
 {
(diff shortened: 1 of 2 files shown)


jaredwray:
@codex review

PR #1826, opened 2026-02-02
sqlite - feat: Add WAL (Write-Ahead Logging) mode support

snomiao:
## Summary

This PR adds Write-Ahead Logging (WAL) mode support to the `@keyv/sqlite` package. Users can now enable WAL mode via the `wal` option in the KeyvSqlite constructor for better concurrency and performance.

## What is WAL Mode?

Write-Ahead Logging (WAL) is an alternative journaling mode in SQLite that provides:
- Better concurrency (readers don't block writers, writers don't block readers)
- Improved performance in many scenarios
- Better crash resistance
- Atomic commits

Reference: https://sqlite.org/wal.html

## Changes Made

1. **Type Definition** (`packages/sqlite/src/types.ts`)
   - Added `wal?: boolean` option to `KeyvSqliteOptions` interface

2. **Core Implementation** (`packages/sqlite/src/index.ts`)
   - Modified the `connect` function to execute `PRAGMA journal_mode=WAL` after database connection if `wal: true` is specified
   - Added proper TypeScript type annotation for Promise<sqlite3.Database>

3. **Tests** (`packages/sqlite/test/test.ts`)
   - Added test to verify WAL mode is enabled when `wal: true` is specified
   - Added test to verify WAL mode is not enabled by default
   - Added test to verify in-memory databases remain in "memory" journal mode even with `wal: true`
   - All tests pass with 100% code coverage

4. **Documentation** (`packages/sqlite/README.md`)
   - Updated README to document the new `wal` option
   - Added Options section explaining all available configuration options
   - Documented the limitation with in-memory databases

## Usage Example

\`\`\`javascript
import Keyv from 'keyv';
import KeyvSqlite from '@keyv/sqlite';

// Enable WAL mode
const store = new KeyvSqlite({
  uri: 'sqlite://path/to/database.sqlite',
  wal: true
});

const keyv = new Keyv({ store });
\`\`\`

### In-Memory Database Limitation

WAL mode does NOT work with in-memory databases (`:memory:`). SQLite automatically uses a special "memory" journal mode for in-memory databases that cannot be changed. This is expected behavior and matches how SQLite works natively - it silently ignores the WAL mode request and remains in "memory" journal mode.

### Default Behavior

WAL mode is **disabled by default** to maintain backward compatibility. Users must explicitly opt-in by setting `wal: true` in the options.

## Benefits of WAL Mode

**Pros:**
- Better concurrency - multiple readers can access simultaneously, writers don't block readers
- Improved performance for most write operations
- Better crash resistance and atomic commits

**Cons:**
- Creates -wal and -shm files alongside the database file
- Not ideal for network filesystems (NFS, SMB)
- Slightly higher disk space usage

## Testing

All tests pass with 100% code coverage:
- ✅ WAL mode is correctly enabled when `wal: true` is specified with file-based databases
- ✅ Default behavior (no WAL) remains unchanged
- ✅ In-memory databases remain in "memory" journal mode (expected behavior)
- ✅ All existing functionality works with and without WAL mode

## References

- [SQLite WAL Documentation](https://sqlite.org/wal.html)
- [SQLite PRAGMA journal_mode](https://sqlite.org/pragma.html#pragma_journal_mode)
- [SQLite In-Memory Databases](https://www.sqlite.org/inmemorydb.html)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -85,3 +85,31 @@ Individual package tests:
 - Core package (`keyv`) depends only on `@keyv/serialize`
 - Storage adapters depend on `keyv` as peer dependency
 - Test suite depends on `keyv` and various testing utilities
+
+## Pull Request Guidelines
+
+### PR Title Format
+Use the following format for pull request titles:
+
+```
+{package} - {type}: {description}
+```
+
+**Examples:**
+- `sqlite - feat: Add WAL (Write-Ahead Logging) mode support`
+- `redis - fix: Connection timeout handling`
+- `keyv - docs: Update API documentation`
+- `mono - chore: Upgrade dependencies`
+
+**Package names:**
+- Use the package name (e.g., `sqlite`, `redis`, `postgres`, `keyv`, `serialize`)
+- Use `mono` for changes that affect the entire monorepo
+
+**Types:**
+- `feat`: New feature
+- `fix`: Bug fix
+- `docs`: Documentation changes
+- `chore`: Maintenance tasks (dependencies, CI, etc.)
+- `refactor`: Code refactoring
+- `test`: Test additions or changes
+- `perf`: Performance improvements


snomiao:
Thanks for the review! I've fixed the documentation discrepancy.

The implementation intentionally matches SQLite's native behavior - when you execute `PRAGMA journal_mode=WAL` on an in-memory database, SQLite silently ignores the request and returns "memory" without any warning or error. This is the expected behavior documented in SQLite's own documentation.

I've updated the README (commit 96e13ff) to accurately reflect this: "SQLite will silently ignore the WAL mode request and remain in 'memory' journal mode."

This approach provides a consistent experience with SQLite's behavior. Users who need to verify the journal mode can query it directly using: `await store.query('PRAGMA journal_mode')`

jaredwray:
@codex review this. focus on anything that is missing

PR #1937, opened 2026-05-29
keyv - fix(v5): listener leak when using once and off

YUCLing:

Bug fix

In v5, a custom `EventManager` is used, and it has a bug that can cause listeners to leak.

https://github.com/jaredwray/keyv/blob/e41ad98e0a9f5629d777a298eb2c20ae92c408af/packages/keyv/src/event-manager.ts#L60-L68

When adding a once listener, it wraps the original inside the newly created `onceListener`, and adds it into listeners. When user code tries to remove it using its original listener, it fails, because `off` cannot find the original listener that is used by user code.

This PR saves the reference of the original listener within the `onceListener` and make `off` also check for it, so it can be removed properly. The PR also changed the max listener warning message, to make it easier to trace.

Example case covered:

https://github.com/jaredwray/cacheable/blob/069f7340194f29f8ad2310cec4b922c0ad21b4f9/packages/cacheable-request/src/index.ts#L316-L325

### DIFF
--- a/packages/keyv/src/event-manager.ts
+++ b/packages/keyv/src/event-manager.ts
@@ -1,5 +1,7 @@
 // biome-ignore lint/suspicious/noExplicitAny: type format
-type EventListener = (...arguments_: any[]) => void;
+type EventListener = ((...arguments_: any[]) => void) & {
+	_originalListener?: EventListener;
+};
 
 class EventManager {
 	_eventListeners: Map<string, EventListener[]>;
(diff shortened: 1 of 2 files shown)


YUCLing:
Hi @jaredwray ! Any particular reason that this is not getting merged? I can work with it!

jaredwray:
@YUCLing - this will get merged. We are having issues publishing to npmjs right now and working with their support so it might take a bit to get this out.

PR #1978, opened 2026-06-17
valkey - refactor: full code and test review

jaredwray:
## Summary

A full code and test review of `@keyv/valkey`, addressing correctness, documentation, and test uniformity with the other storage adapters (modeled on `@keyv/redis`).

## Source changes (`src/index.ts`, `src/types.ts`)

- **Return `undefined`, never `null`** — `getMany()` now maps missing keys to `undefined` in both the standalone and cluster paths, and `iterator()` yields `undefined` for missing values (previously raw `MGET` `null`s leaked through).
- **Namespace is fully supported natively** — the constructor now honors a `namespace` option (added to `KeyvValkeyOptions`), and `createKeyv()` propagates it to the store, so namespacing works end to end (constructor option, setter, and Keyv wrapping).
- **Events wired correctly via hookified** — extracted a single private `initClient()` that re-emits `error`/`connect`/`reconnecting`, guards against duplicate listeners, and is also called from the `client` setter so a replaced client keeps forwarding errors (previously the setter left the new client unwired).
- **Robust JSDoc & types** — fixed incorrect return types/descriptions on `set` (`Promise<boolean>`), `setMany` (`Promise<boolean[] | undefined>`), and `iterator`; removed the non-existent `namespace` parameter from `iterator()`'s docs; added explicit return types to `delete`, `clear`, `has`, and `disconnect`.
- **Member ordering** — properties (getters/setters) sit directly under the constructor and above the methods; all private functions are below the public ones.

## Tests

- Split the monolithic `test/test.ts` into per-feature files — `get`, `set`, `delete`, `has`, `namespace`, `iterator`, `events`, `main`, and `suite` — using `describe`/`test`/`expect` with `faker` and the shared `delay` helper, matching the other storage modules.
- Added explicit coverage for the new behaviors: `getMany` → `undefined`, native constructor namespace, `createKeyv` namespace propagation, event re-wiring, and `iterator()` used without a namespace argument.
- Cleaned up `cluster.test.ts` to the same `describe`/`test`/`expect` style.
- Run suites sequentially (`fileParallelism: false`) to avoid cross-file interference on the shared Valkey instance.

## Docs (`README.md`)

- Documented the `namespace` constructor option.
- Removed the reference to a non-existent `opts` getter from the migration guide.
- Fixed variable shadowing (`const KeyvValkey = new KeyvValkey(...)`, a TDZ `ReferenceError`) in the usage examples and corrected the `LICENSE` link.

## Verification

`pnpm build`, `biome check`, and the full non-cluster suite (123 tests across 9 files) all pass locally against a Valkey-compatible server. Cluster tests are unchanged in behavior and run in CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
_Generated by [Claude Code](https://claude.ai/code/session_018Jkvomk1WeYnaa7mHmCdLG)_


jaredwray:
## Review feedback addressed

Thanks for the automated reviews — here's what was done for each item.

**🟥 Gemini — `getMany([])` protocol error (`src/index.ts`)**
Confirmed `MGET` with no keys throws `ERR wrong number of arguments for 'mget' command`. Added an early `if (keys.length === 0) return [];` guard at the top of `getMany` (matching the redis adapter), which also short-circuits the cluster path. Regression test added in `get.test.ts`. — `0413b1a`

**🟥 Gemini — stable listener refs + cleanup on client replacement (`src/index.ts`)**
Moved the `error`/`connect`/`reconnecting` handlers to `readonly` class-field references (`_errorHandler`, `_connectHandler`, `_reconnectingHandler`). `initClient()` now `removeListener`s them from the previous client before wiring the new one, so replacing `client` no longer leaks listeners or re-emits from a discarded client. The events test now asserts the old client's listener count drops to 0. — `0413b1a`

**🟨 Codex (P2) — `createKeyv` namespace erased via options object**
Good catch. `createKeyv({ uri, namespace })` left `options` undefined, so the `Keyv` wrapper was built with `namespace: undefined`, which `Keyv.setStore` wrote back over the adapter's resolved namespace. Fixed by deriving the wrapper namespace from `adapter.namespace` (which already resolves both the connect-object and options-argument forms). Added a regression test asserting keys land under `namespace:<ns>:<key>` and not at the bare key. — `a1f3d24`

**Codecov — patch/project back to 100%**
Refactoring the handlers to stable references left the `reconnecting` body as the only uncovered line (it never fires in tests; the inline form had been counted via the `.on()` call). Added deterministic tests that emit `connect` and `reconnecting` on the client and assert the adapter re-emits them. — `a1f3d24`

All 14 checks are green on the latest commit and project coverage is 100%.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2040, opened 2026-08-03
test-suite - fix: de-flake storage TTL expiry tests

jaredwray:
- [x] Tests for the changes have been added (for bug fixes/features) with 100% code coverage. _(Test-only change — no production code is touched; the shared `@keyv/test-suite` TTL tests are made deterministic. Verified the full `@keyv/sqlite` suite green on Node 22 and Node 24 and the `@keyv/test-suite` unit tests green.)_

Bug fix — removes intermittent flakiness in the shared storage TTL test suite.

### The problem

`storage/sqlite test:ci` intermittently failed on Node 24:

```
FAIL  test/test.ts > setMany with per-entry expires expires individual entries
AssertionError: expected undefined to be 'vito' // Object.is equality
- Expected: "vito"
+ Received: undefined
  ❯ ../../core/test-suite/dist/index.mjs:683:31   // t.expect(await s.get(key1)).toBe(val1)
```

The same flake was also observed on the MySQL adapter, which is the tell that this is **not** a SQLite or Node 24 bug — it lives in the shared suite every adapter runs (`core/test-suite/src/storage-ttl.ts`).

### Root cause

The affected tests capture an **absolute** expiry deadline (`Date.now() + ttl`) as an argument *before* the write executes, then immediately assert the value is still retrievable:

```ts
await s.setMany([{ key: key1, value: val1, expires: expiresIn(ttl) }, ...]); // deadline fixed BEFORE the write
t.expect(await s.get(key1)).toBe(val1);                                       // races against that deadline
```

Because the deadline is fixed up front, **any latency between capturing it and the read-back counts against `ttl`**. The first operation on a freshly created store also pays for connection setup (cold connect, auth, schema check, initial lock acquisition). Under CI load — coverage instrumentation, parallel vitest workers contending on the shared file DB, networked backends like MySQL — that first-operation cost can exceed the very tight `100ms` `ttl`. When it does, `get` sees the deadline already in the past, deletes the row, and returns `undefined` — exactly the reported failure.

This was proven deterministically: injecting ≥ `ttl` of latency between capturing the deadline and reading it back reproduces `undefined` every time.

### The fix

Two complementary changes in `core/test-suite/src/storage-ttl.ts`:

1. **Warm up the store's connection before the deadline is captured** (a throwaway `get`), so connection-setup latency is no longer counted against the TTL window. This is the dominant, adapter-variable cost and the main reason the flake was load-dependent and cross-adapter.
2. **Widen the millisecond-granularity window** so ordinary write+read latency cannot consume it: `ttl` `100ms → 500ms` and `expiryDelay` `200ms → 1000ms`. The second-granularity path (`1000ms`/`3000ms`) is unchanged.

Measured against the real `@keyv/sqlite` suite on Node 24, with coverage and the parallel `drivers.test.ts` contending on the same file DB, the write+read now completes in **~5ms against the 500ms window** — roughly a 100× safety margin.

### Verification

- `@keyv/sqlite` full suite (coverage, `--sequence.setupFiles=list`): **149/149 passed** on both Node 24 and Node 22.
- `@keyv/test-suite` unit tests (which exercise `storageTtlTests` at both granularities): **93/93 passed**.
- `biome check` clean on the changed file.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01SUUxKcgcxUr8FDtx2RzgGb

---
_Generated by [Claude Code](https://claude.ai/code/session_01SUUxKcgcxUr8FDtx2RzgGb)_

### DIFF
--- a/core/test-suite/src/storage-ttl.ts
+++ b/core/test-suite/src/storage-ttl.ts
@@ -25,8 +25,14 @@ const storageTtlTests = (test: TestFunction, store: StorageFn, options?: Storage
 	// the entry to still be live. Keep `ttl` comfortably larger than a single write+read
 	// against a store under CI load (coverage instrumentation, lock contention, networked
 	// backends) so the entry has not already expired by the time it is first read.
-	const ttl = seconds ? 1000 : 500;
-	const expiryDelay = seconds ? 3000 : 1000;
+	//
+	// Both millisecond-path values must stay strictly sub-second (expiryDelay < 1000ms). A
+	// backend that only honors whole-second TTLs rounds these deadlines up to ~1s, so waiting
+	// a full second before the expiry check would let its key expire and hide that mismatch.
+	// Staying under a second keeps the millisecond suite failing for such backends, which is
+	// the signal for them to opt into `ttlGranularity: "seconds"`.
+	const ttl = seconds ? 1000 : 300;
+	const expiryDelay = seconds ? 3000 : 600;
 	// The storage contract takes an absolute expiry (ms since epoch), not a relative ttl.
 	const expiresIn = (ms: number) => Date.now() + ms;
 
--- a/core/test-suite/src/storage-ttl.ts
+++ b/core/test-suite/src/storage-ttl.ts
@@ -20,13 +20,29 @@ const storageTtlTests = (test: TestFunction, store: StorageFn, options?: Storage
 
 	const missingValue = options?.missingValue;
 	const seconds = options?.ttlGranularity === "seconds";
-	const ttl = seconds ? 1000 : 100;
-	const expiryDelay = seconds ? 3000 : 200;
+	// The absolute `expires` deadline is captured (as `Date.now() + ttl`) before the write
+	// runs, so the write and the immediate read-back both have to complete within `ttl` for
+	// the entry to still be live. Keep `ttl` comfortably larger than a single write+read
+	// against a store under CI load (coverage instrumentation, lock contention, networked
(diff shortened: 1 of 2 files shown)


core/test-suite/src/storage-ttl.ts
```diff
@@ -20,13 +20,35 @@ const storageTtlTests = (test: TestFunction, store: StorageFn, options?: Storage
...
+	// runs, so the write and the immediate read-back both have to complete within `ttl` for
+	// the entry to still be live. Keep `ttl` comfortably larger than a single write+read
+	// against a store under CI load (coverage instrumentation, lock contention, networked
+	// backends) so the entry has not already expired by the time it is first read.
+	//
+	// Both millisecond-path values must stay strictly sub-second (expiryDelay < 1000ms). A
+	// backend that only honors whole-second TTLs rounds these deadlines up to ~1s, so waiting
+	// a full second before the expiry check would let its key expire and hide that mismatch.
+	// Staying under a second keeps the millisecond suite failing for such backends, which is
+	// the signal for them to opt into `ttlGranularity: "seconds"`.
```
jaredwray: this should be 20% the size

jaredwray: Trimmed to 3 comment lines (~19% of the original 16) in f8fd7c9 — kept only the two essentials: the deadline is captured before the write, and the ms path must stay sub-second so second-only backends still fail it.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #2052, opened 2026-08-19
aws-ssm - feat: Add AWS SSM (Systems Manager Parameter Store) storage adapter

DManavi:
## Summary

Adds `@keyv/aws-ssm`, a new Keyv storage adapter backed by **AWS Systems Manager Parameter Store**, built on **AWS SDK v3** (`@aws-sdk/client-ssm`).

## Key design decisions

- **The `SSMClient` must be provided by the caller.** The adapter never constructs or owns a client itself (unlike `@keyv/dynamo`), and never calls `.destroy()` on it — Parameter Store access typically needs application-managed credentials/config (IAM roles, profiles, STS, etc.), and the client may be shared elsewhere in the consumer's app.
- **Parameter names are always prefixed** — `keyPrefix` defaults to `/keyv/` (normalized to always start/end with `/`), with an optional `namespace` inserted after it. This also acts as a safety boundary: `clear()`/`iterator()` only ever touch parameters under the configured prefix, never anything else in the account.
- **Client-side TTL enforcement.** Parameter Store has no reliable native per-parameter expiry (Advanced-tier "Expiration" policies are coarse/async), so values are stored as a small JSON envelope (`{ v, e }`) and expiry is enforced on every read, with lazy deletion of expired parameters — the same technique already used by `@keyv/etcd`.
- **Configurable `type`** (`String` / `StringList` / `SecureString`) **and `tier`** (`Standard` / `Advanced` / `IntelligentTiering`), with a `keyId` option for KMS-encrypted `SecureString` values. Defaults to `String`/`Standard` (AWS's own defaults, zero-config, no KMS permissions required out of the box).
- **Batching matches real AWS limits**: `getMany`/`hasMany`/`deleteMany` chunk into groups of 10 (the hard limit for `GetParameters`/`DeleteParameters`); `setMany` runs parallel individual `PutParameter` calls since AWS has no batch put API for SSM; `clear()`/`iterator()` paginate through `GetParametersByPath` with a manual `NextToken` loop (rather than the SDK's `paginateGetParametersByPath` helper, which does a strict `instanceof SSMClient` check that would reject test doubles and some wrapped/proxied clients).

## Testing

AWS has no official local SSM emulator. Per discussion, this PR uses **fully mocked/offline tests** rather than LocalStack/Docker: a small in-memory fake `SSMClient` (`test/helpers/ssm-mock.ts`, implementing only `.send()` with real `@aws-sdk/client-ssm` command/exception classes for fidelity, including the real 10-item batch caps and `NextToken` pagination) drives:
- the full shared `@keyv/test-suite` compliance suite (`keyvTestSuite`, `keyvIteratorTests`, `storageTestSuite`)
- adapter-specific tests: construction/validation, namespace/`keyPrefix` formatting and double-prefix avoidance, `type`/`tier`/`keyId` options, batching/pagination beyond the 10-item AWS limit, TTL/expiry, error handling, and raw/non-enveloped values.

**122/122 tests passing, 100% statement/branch/function/line coverage on `src/index.ts`.**

No Docker services or CI workflow changes were needed since tests run fully offline.

## Docs

- New `storage/aws-ssm/README.md`, mirroring the structure of `storage/dynamo`/`storage/etcd` (features, quick start, namespace usage, `SecureString` usage, options/properties/methods reference, and Parameter Store limits).
- Root `README.md` updated: added `aws-ssm` to the Project Structure tree and Storage Adapters list.

## Checklist

- [x] `pnpm build` (keyv core → aws-ssm) succeeds
- [x] `pnpm test` / `pnpm test:ci` (biome lint + vitest coverage) passes for the new package
- [x] Root README updated
- [x] No changes needed to `pnpm-workspace.yaml`, CI workflows, or Docker test services

🤖 Generated with [Copilot CLI](https://github.com/github/copilot-cli)


jaredwray:
@DManavi - Hi! I have been doing research on this and I don't think it makes sense to add this to the core repo as we are being more picky about what are are adding here. One thing you could do is publish this yourself and list it as a third party adapter instead.

DManavi:
Sure, thanks, @jaredwray. What's the process of adding it to the 3rd-party adapters list on keyv website?

Issue #1566, opened 2025-09-04
Support for Valkey Glide client


ofek29:
I’d like to suggest adding support for Valkey Glide client in Keyv.
Valkey Glide is an open-source project, designed for reliability and performance, and it supports Valkey, and Redis open source.
I’d be happy to help with the implementation if this is something the project would accept.

jaredwray:
@ofek29 - sounds interesting and I would be happy to look at doing this and getting some help on it. One thing of note is that we are working to do a full rewrite on `@keyv/valkey` in October 2025 so you might want to wait for that. Thoughts?

Also, do you have a couple of links to glide so I can learn more?

ofek29:
I think the best option is to have two separate packages, iovalkey and Glide. As they are different in implementation and setup and both are very popular.
Also important to consider that Glide is built with Rust core and bindings.
Valkey-Glide: [github](https://github.com/valkey-io/valkey-glide), [docs](https://valkey.io/valkey-glide/)

jaredwray:
@ofek29 - after looking at this I am wondering if we can put it in a single package with ioredis as it looks like you can create Glide clients which are unique. The default would be iovalkey but you can then add in Glide as the client if you want and it works with both? I would prefer to only support one package for valkey if possible.

avifenesh:
Hi :)
I'm Valkey GLIDE maintainer. 
Just want to add a few points, pretty neutrals.
iovalkey support is easier to give, as it is a drop-in for ioredis; it is actually better as well for redis users to use iovalkey since it is compatible and has some degree of maintenance, while ioredis is kind of dead.
It is easier for users who already use ioredis to drop ioredis and drop in iovalkey since it is fully compatible.
As a Valkey community member, I would say that I would love to see iovalkey supported, and since it's easier, probably first.

On the other hand, going over the API, I see “delete many,” “set many,” and “get many.”
Generally, when using valkey, you would like to perform those many in one round trip, using mset, mget, and del.
If you are using valkey cluster, you can't use it with cross-slot keys, but GLIDE supports those operations and sends MGET, MSET, DEL, and FLUSHALL as cluster-aware operations and handles them transparently.

So the implementation under the hood for dual mode is not so smooth in case you want to provide the best to your users. In addition, if I'm using GLIDE, I will be less thrilled to have a built-in unused dependency by default.

Anyway, if my help is needed for anything in the key value area, especially GLIDE, please let me know; I would be super happy to help.

@jaredwray @ofek29

[...]

PR #2146, opened 2026-09-10
valkey-glide - feat: Add Valkey GLIDE storage adapter

EmilBuszylo:
Adds `@keyv/valkey-glide` so Keyv can use the official [Valkey GLIDE](https://glide.valkey.io/) Node client (`@valkey/valkey-glide`). This is the path discussed in #1566: keep `@keyv/valkey` on `iovalkey`, and land GLIDE as a separate storage adapter.

Feature: new official storage adapter `@keyv/valkey-glide`.

## Why

GLIDE is the Valkey client with a Rust core, cluster-aware multi-key commands (`MGET`, `UNLINK`, …), and read strategies including **AZ affinity** (`readFrom` + `clientAz`). That is not a drop-in for `iovalkey`, so a dedicated package matches the maintainer note on #1566.

## What

- New package `storage/valkey-glide` implementing the v6 `KeyvStorageAdapter` contract (`expires` via `SET` + `PXAT`).
- Lazy connect: `GlideClient.createClient` / `GlideClusterClient.createClient` are async; the first command or `getClient()` opens the connection. Existing GLIDE clients can be passed in.
- Constructor accepts a URI (`redis://`, `rediss://`, `valkey://`, `valkeys://`), a GLIDE config object (`cluster`, `addresses`, `readFrom`, `clientAz`, `useTLS`, …), or an existing client.
- `getMany` uses GLIDE `mget` (cluster-aware). `clear()` / `iterator()` use `SCAN` (cluster `ClusterScanCursor` when needed).
- Same namespace / `useSets` key layout as `@keyv/valkey`. `useSets: true` is documented as not cluster-safe.
- Docs: package README, root / `keyv` README, website adapter overview.

AZ affinity example:

```js
const store = new KeyvValkeyGlide({
  cluster: true,
  addresses: [{ host: 'clustercfg.example.cache.amazonaws.com', port: 6379 }],
  useTLS: true,
  readFrom: 'AZAffinity',
  clientAz: 'us-east-1a',
});


jaredwray:
@EmilBuszylo - thanks for doing the work on this. Here is what we need to do before we merge this

## Blocking

1. **Fix `setMany` and `deleteMany` to use GLIDE batches.** Both are `Promise.all` over per-key commands (`storage/valkey-glide/src/index.ts:192` and `:217`), which exceeds GLIDE's default in-flight limit of 1000. Use `Batch` for standalone and `ClusterBatch` for cluster via `client.exec(batch, raiseOnError)`, one SET with PXAT per entry plus SADD when `useSets` is on, and map results back to the input order. Same for `delete()` with `useSets` (line 206), which currently fires two parallel commands per key.
2. **Fix the `clear()` pattern.** Line 266 uses `${prefix}*`, so namespace `foo` also wipes `foobar`. Change to `${prefix}:*` and add a test with two namespaces that share a prefix.
3. **Make `iterator()` stream.** Lines 273 to 288 collect every key with SCAN and then issue a single MGET for all of them. Yield per SCAN page with one MGET per page, in both the standalone and cluster paths. Optionally do the same paging for `clear()` so it never issues one giant UNLINK.
4. **Bring coverage to 100% and cover the headline features.** Uncovered today: the whole cluster SCAN path (lines 353 and 360 to 373), option pass-through (line 421, so `readFrom` and `clientAz` are never exercised), URI credentials and database parsing (lines 448 and 456 to 458), the connect-failure path (lines 313 to 314), `useSets` with no namespace (lines 323 and 332), namespace with a passed-in client (line 62), and the Buffer branches (lines 480 to 484 and 508). Cluster tests should call `clear()` and `iterator()` and should assert the `{ cluster: true, addresses }` constructor path works.
5. **Set both new `allowBuilds` entries to `false`** in `pnpm-workspace.yaml`. Neither package needs its scripts and install still passes.
6. **Correct the PR checklist.** The 100% coverage box is checked but the suite measures 88% statements and 74% branches.

## Requested, non-blocking

7. **Tighten the dependency range** from `^2.2.0` to `^2.5.2`. Only 2.5.2 was tested, and the batch API is what the fix in item 1 relies on.
8. **README: document GLIDE defaults.** Request timeout is 250 ms, connection timeout is 2 s, in-flight limit is 1000. Show how to override them through the pass-through options.
9. **README: state platform support.** Linux (glibc and musl) and macOS only, no Windows, native binary, roughly 20 MB installed.
10. **README: reword the `useSets` caveat.** It was copied from `@keyv/valkey`. This adapter has no MULTI, so `useSets` works on a cluster (verified). It is non-atomic, which is the accurate warning.
11. **Emit `error` once on connect failure.** `createClient` emits and rethrows (line 313), then `set()` catches and emits again, so one failed write produces two events.
12. **Remove the synchronous `connect` emit in the constructor** (line 65). No listener can exist yet, so it never reaches anyone.
13. **Pass Buffers through instead of stringifying.** `toGlideValue` (line 475) converts `Uint8Array` with `toString()`, which replaces invalid UTF-8 bytes. GLIDE accepts `Buffer` directly. Low impact because Keyv serializes first, but the current behavior is silently lossy.
14. **Make `hasMany` use EXISTS.** It currently fetches full values through `getMany` (line 231) just to test presence. A batch of EXISTS avoids moving large values.
15. **Harden `parseConnectionUri`.** `decodeURIComponent` on a malformed password throws `URIError` (line 450), a username with no password yields an empty-string password, and query parameters are silently ignored. Either handle these or document them.

